Issue #1, opened 2026-05-06
Fix live bandit attack visuals


claude-do:
## Summary
Make the existing bandit marker, countdown, pulse, and short attack vignette work when VITE_CLANWORLD_DEMO_MODE=false.

## Acceptance Criteria
- Live snapshot exposes active bandit data from Convex banditView.
- WorldMap renders active bandit visuals in live mode.
- BanditAttackResolved chain events trigger the existing short combat vignette with success/failure outcome.
- Demo mode tick-48 behavior remains intact.
- Server and web typechecks pass.

PR #3, opened 2026-05-06
fix(web): animate live bandit attacks

claude-do:
## Summary
- exposes active bandit state through the snapshot query
- renders bandit marker/countdown/pulse in live mode instead of only demo mode
- triggers the existing short combat vignette from live `BanditAttackResolved` events

## Why
Real bandit raids were visible in the ticker, but the Pixi animation was still gated behind `VITE_CLANWORLD_DEMO_MODE=true` and the hardcoded demo tick 48.

Closes #1

## Validation
- `pnpm --filter @clan-world/server typecheck`
- `pnpm --filter @clan-world/web typecheck`
- `VITE_CLANWORLD_DEMO_MODE=false pnpm --filter @clan-world/web build`

### DIFF
--- a/apps/server/convex/getSnapshot.ts
+++ b/apps/server/convex/getSnapshot.ts
@@ -6,6 +6,7 @@ import {
   WINTER_PERIOD_TICKS,
   SEASON_DURATION_TICKS,
 } from "@clan-world/shared/generated/constants";
+import { BanditState } from "@clan-world/shared/generated/enums";
 
 const SEASON_LEN = Number(SEASON_DURATION_TICKS);
 const WINTER_START = Number(WINTER_START_TICK);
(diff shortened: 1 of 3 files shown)


claude-do:
**Round-2 fixes (commit `096331d`)**

### ✅ Addressed
- HIGH `WorldMap.tsx:2928` — loop break added, only the first unseen `BanditAttackResolved` event triggers per render; remaining unseen events stay un-`seen`'d and pick up next render
- MEDIUM `WorldMap.tsx:2905` — magic number `15_000` → named constant `RECENT_BANDIT_EVENT_THRESHOLD_MS` hoisted to module level with intent comment
- MEDIUM `WorldMap.tsx:1000` — overly complex `(snapshot as { bandit?: ... })` cast simplified to `snapshot?.bandit ?? null`
- MEDIUM `TopHud.tsx:112` — same cast simplified

### ⏭️ Deferred (code-quality, not bugs)
- MEDIUM `WorldMap.tsx:2196` — `pointertap` handler dedup between `banditIcon` and `sprite`. Worth a follow-up but bigger touch; leaving for a dedicated refactor pass

`tsc --noEmit` clean.

PR #7, opened 2026-05-06
fix(web): smooth base breathing animation (#5)

### LINKED ISSUE #5: Fix choppy base bobbing animation (subpixel snap bug)
## Bug

Base sprites on the world map "jump" between three pixel rows instead of bobbing smoothly like clansmen do.

## Root cause

`apps/web/src/WorldMap.tsx:1371`:

\`\`\`js
base.container.y = base.baseY + Math.round(Math.sin((now + base.phaseOffset) * Math.PI / 2000) * 1);
\`\`\`

`Math.sin(...)` returns a value in `[-1, 1]`, multiplied by `1`, then `Math.round`'d → output snaps to only `-1`, `0`, or `+1`. The base teleports between three rows once per quadrant of the cycle.

Compare clansman bob at `WorldMap.tsx:2096` (smooth, no `round`):
\`\`\`js
const bob = Math.sin(performance.now() / 320 + marker.offsetIndex * 1.7) * 1.2;
\`\`\`

## Fix

Replace the bob with a **breathing** animation: scaleY stretch from a fixed bottom anchor (no Y translation). Keeps bases planted on the ground and reads as "alive" rather than "jumping."

Plan:
- Anchor base sprite pivot to bottom-center so scaleY grows upward only.
- Replace the per-frame y write with a per-frame scaleY write (e.g. `1 + sin(...) * 0.03`).
- Drop Math.round — Pixi handles subpixel rendering.
- Keep the existing phaseOffset so bases breathe out of sync.

claude-do:
## Summary

- Bases on the world map jumped between three pixel rows instead of moving smoothly. The bob at `WorldMap.tsx:1337` wrapped `Math.sin(...) * 1` in `Math.round`, snapping output to `{-1, 0, +1}`.
- Replaced the y-translation with a **breathing** animation: `scale.y` stretch from the sprite's bottom anchor (`0.5, 1` — already set), 3% amplitude. Reads as alive without skewing the painted rooflines.
- Dropped `Math.round` so Pixi renders subpixel. Kept `phaseOffset` so bases breathe out of sync.

## Test plan

- [ ] `pnpm --filter @clan-world/web typecheck` — passes locally
- [ ] Visually verify bases breathe smoothly on the world map (no Y jitter, no roofline skew)
- [ ] Confirm `phaseOffset` still desyncs neighboring bases

Closes #5

### DIFF
--- a/apps/web/src/WorldMap.tsx
+++ b/apps/web/src/WorldMap.tsx
@@ -1332,12 +1332,16 @@ export function WorldMap() {
           const now = performance.now();
           drawn.bases.forEach((base) => {
             if (base.baseY <= 0) return;
-            base.container.zIndex = Math.round(base.baseY);
             // Breathing: scaleY stretches upward from the sprite's bottom anchor
             // (0.5, 1). 0.25 Hz period; 3% amplitude reads as alive without skewing
             // the painted rooflines. phaseOffset keeps bases out of sync.
+            // zIndex is already set during relayout (base.baseY is static between
+            // relayouts), so no need to reassign each tick.
+            const scaleY = 1 + Math.sin((now + base.phaseOffset) * Math.PI / 2000) * 0.03;
             if (base.sprite) {
-              base.sprite.scale.y = 1 + Math.sin((now + base.phaseOffset) * Math.PI / 2000) * 0.03;
+              base.sprite.scale.y = scaleY;
+            } else {
+              base.fallback.scale.y = scaleY;
             }
           });
           updateLiveClansmanPositions();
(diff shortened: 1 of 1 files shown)


claude-do:
**Round-2 fix (commit `029f93c`)**

### ✅ Addressed
- MEDIUM `WorldMap.tsx:1341` — both findings:
  - Dropped per-tick `base.container.zIndex` reassignment in `zonePulseCb`. zIndex is already set during relayout at `WorldMap.tsx:2384` and `base.baseY` is static between relayouts.
  - Breathing animation now applies to `base.fallback` when `base.sprite` is null. Previously the fallback sat statically while sibling bases breathed.

`tsc --noEmit` clean.

PR #148, opened 2026-05-10
feat(contracts): bandit 3-tick attack cycle (remove Escaped + Resting states)

claude-do:
Per Liam directive 2026-05-10 — bandit recovery cycle was 8 ticks between attacks (Escaped 2 + Resting 2 + Camped 3 + Attacking instant). Target was 3 ticks. This PR removes the `Escaped` and `Resting` states entirely so the cycle becomes `Camped (3) → Attacking (instant) → Camped (3, next region) → Attacking ...`.

## State machine

```
Spawned (1 tick) → Camped (3 ticks) → Attacking-instant
                                            │
                                            ├── Defeated (terminal)
                                            │
                                            └── Escape outcome → Camped (3, NEXT region) → ...
```

## Cadence
- First attack resolves at end of tick **4** (1 Spawned + 3 Camped)
- Subsequent attacks at end of ticks **7, 10, 13...** (3-tick cycle)
- After 6 attempts: terminal escape (`BANDIT_MAX_ATTACK_ATTEMPTS`)

## Changes (12 files, +178 / −235)

- **`packages/contracts/src/IClanWorld.sol`**: `BanditState` enum reduced to `{None, Spawned, Camped, Attacking, Defeated}` (was 7 values)
- **`packages/contracts/src/diamond/lib/LibBanditLifecycle.sol`**: Drop `Escaped → Resting` and `Resting → Camped` clauses. Move-to-next-region helper now fires on `Attacking → Camped` (escape outcome) instead of Resting → Camped.
- **`packages/contracts/src/diamond/lib/LibBanditCombat.sol`**: On winning attack, transition `Attacking → Camped` directly + move bandit to next rampage region.
- **`packages/contracts/src/ClanWorldLens.sol`** + **`BanditViewsFacet.sol`** + **`ClanWorld.sol`** (legacy monolith): Remove Resting/Escaped branches from view paths.
- **`packages/shared/src/generated/enums.ts`**: BanditState enum regen.
- **Tests**: `Bandit.t.sol`, `BanditAttackResolution.t.sol`, `DiamondSkeleton.t.sol` updated to assert new states.
- **Docs**: `clanworld_v4_6_bandit_phase9_redesign.md` state machine + `clanworld_v5_animation_spec.md` combat phase timing rewritten.

## Verification

- **73/73 targeted tests pass**: BanditTest (13) + BanditAttackResolutionTest (23) + DiamondSkeletonTest (37).
[...]

PR #150, opened 2026-05-10
feat(web): bandit attack animation — telegraph + battle + outcome variants

claude-do:
## Summary

Implements bandit attack animation per [`docs/planning/bandit-animation-impl-plan.md`](https://github.com/clan-world/clan-world-game/blob/docs/bandit-animation-plan/docs/planning/bandit-animation-impl-plan.md).

Stacked PR — base is `feat/bandit-3-tick-cycle` (PR #148). Merge order: #148 first, then this.

## Animation flow

- **T3 last 10s**: camp morphs to 3 standing bandits (telegraph). Jitter ±1px, phase-offset.
- **T4 full 60s**: full battle animation, post-resolution. Outcome derived from snapshot diff:
  - **defeat**: march → circle → flash → 3-frame death + tombstone fade
  - **win**: converge + base shake → cluster pause → walk to next region → glow camp
  - **no-target advance** (forward-compatible with stacked backend PR `feat/bandit-no-target-advances`): walk to new region without battle
  - **terminal escape**: walk off map edge

## Sprite kit (Liam shipped 2026-05-10)

Sliced from raw strips at `/tmp/bandit-sprites-staged/` into per-frame PNGs:
- 4-frame NE walk + horizontal mirror = NW
- 4-frame SE walk (top row of 2-row source) + horizontal mirror = SW
- 3-frame death sequence (back → face-down → tombstone)
- Camp sprite is opportunistically loaded (`bandit_camp.png` if present, falls back to existing `/sprites/bandit.png`).

## Implementation notes

- **Snapshot-diff outcome tracking**: `prevBanditRef` + `lastBanditOutcomeRef` derive outcome from the snapshot transition. Frontend cannot pre-compute the winner — outcome is observed end-of-tick after the heartbeat closes resolution.
- **`computeBanditAnimPhase`** is a pure function (exported for unit testing) returning one of `camp_idle`, `camp_telegraph`, `battle`, `no_battle_advance`, `terminal_escape`, or `hidden`.
- **Animation clock independent of tick clock**: 8 FPS walk frame indexer via `Date.now()`; tick-based phase boundaries via `currentTickFloat`.
- **DEMO_MODE preserved**: legacy `?combat=success` vignette path still triggers via the existing combat-vignette code, untouched. Phase machine activates only when a live snapshot is present.
- **No changes to contracts** (per brief). The backend `feat/bandit-no-target-advances` PR is forward-compatible — heuristic on `prev.projectedTargetClanId` selects `won` vs `no_battle_advance` from the diff.

## Test plan
- [ ] Visual UAT: camp renders during T0–T2.83, glow ramps
- [ ] Visual UAT: telegraph (3 standing bandits) appears T2.83
- [ ] Visual UAT: defeat scenario plays full sequence
- [ ] Visual UAT: win scenario rampages to next region
- [ ] Reload during mid-T4: animation phase resumes correctly
- [ ] No regressions on existing combat vignette behavior (DEMO_MODE `?combat=success`)

Verification at HEAD:
- `pnpm --filter @clan-world/web typecheck` ✅
- `pnpm --filter @clan-world/shared typecheck` ✅
- `pnpm --filter @clan-world/web lint` ✅ (stub)
- `pnpm --filter @clan-world/web build` ✅


mcorrig4:
@codex please review

claude-do:
# Phase Super-Swarm Synthesis — PR #150 (head d26e937)

**Models run:** Codex 5.4 ✓ | Codex 5.5 ✓ | Opus 4.6 ✗ (pending — 0 bytes at synthesis time) | Opus 4.7 ✓ | Gemini 3.1 Pro ✓
**PR:** feat(web): bandit attack animation — telegraph + battle + outcome variants
**Reason:** Bandit attack animation impl — new render phase machine in WorldMap.tsx with snapshot-diff outcome tracking. Frontend-heavy PR with animation correctness, mid-tick reload behavior, and forward-compatibility with backend stacked PR #151.
**Diff size:** 852 lines

---

## Summary

**Verdict: NEEDS_FIXES** — 4 distinct HIGH-severity findings across reviewers, with 2 at 3/4+ consensus. The most critical (Opus 4.7 unique find): the new T4 battle phase is effectively dead code in live production because the legacy chain-event combat vignette still preempts the new phase machine. The second critical cluster (3/4 consensus): `computeBanditAnimPhase()` evaluates the live bandit before checking `lastOutcome`, so `won` and `no_battle_advance` animations never play — the bandit teleports to the new region instead. Cold-start T4 reload is also broken (2 HIGH + 1 MED consensus). The PR is one targeted gating change away from working in live mode, but the stale-outcome edge cases need fixes too before ship.

**Recommended action:** Path A fix-round. The H1 (legacy vignette preempt) fix is a single `useEffect` gate change. H2/H3 (phase machine priority + Camped-only guard) are small logic inversions. H4 (cold-start) may be deferred if acknowledged as a known limitation. Scope is ~2-3h codex.

---

|---|---|---|---|---|
| H1 | `WorldMap.tsx:3840` | 1/4 (Opus 4.7 unique — but HIGH impact) | HIGH | Legacy `BanditAttackResolved` chain-event useEffect fires in live mode (not DEMO_MODE only), pushing to `pendingCombatTriggerRef` → `startCombatVignette` runs → new phase container goes `visible=false` for the entire T4. Every code path inside `renderBattlePhase` (march/circle/flash/death/won-walk) is unreachable in production. Fix: gate the chain-event useEffect behind `DEMO_MODE` OR invert the container-yield check so the new phase takes over in live mode. |
| H2 | `WorldMap.tsx:876` / `computeBanditAnimPhase` | 3/4 (Codex 5.5 HIGH, Codex 5.4 implied by guard analysis, Gemini HIGH) | HIGH | `computeBanditAnimPhase()` evaluates live `bandit.state` before checking `lastOutcome`. When a bandit wins or advances to a new region, it appears in `snapshot.bandit` as `Camped` immediately, causing `camp_idle` to return before the `lastOutcome` transition window is checked. `won` and `no_battle_advance` 60s animations never play — bandit teleports to new region. Fix: evaluate `lastOutcome` first, with a time-window guard (`currentTickFloat - lastOutcome.resolvedTick < BATTLE_T_FADE_END`), before falling through to steady-state live bandit rendering. |
| H3 | `WorldMap.tsx:1228` | 2/4 HIGH + 1/4 MED (Codex 5.4 HIGH, Codex 5.5 HIGH, Gemini MED) | HIGH | Cold-start (page reload during T4): `prevBanditRef.current` is null on first render, so the diff useEffect never fires, `lastBanditOutcomeRef` is never populated. User who reloads mid-T4 sees either `camp_telegraph` frozen or immediate snap to post-resolution state with no battle animation. Fix: either persist outcome data in backend snapshot payload, or explicitly acknowledge and document this as a known gap. |
| H4 | `WorldMap.tsx:1228` | 1/4 HIGH + 1/4 MED (Codex 5.4 unique HIGH, Opus 4.7 MED as stale-outcome variant) | HIGH | Diff watcher guards on `prev.state === BanditState.Camped`. If the indexer ever exposes a snapshot with `state=Attacking` (or `state=Defeated`) before bandit is null, the outcome derivation is skipped entirely — `lastBanditOutcomeRef` never set — and the renderer stalls at telegraph or goes hidden. Widening the guard to also handle `prev.state === Attacking` (when `tick >= prev.nextActionTick`) closes this. |

---

## SHOULD FIX (MEDIUM)

| # | Models | Finding |
|---|---|---|
| M1 | 4/4 (all reviewers) | `terminal_escape` is dead code. Diff path always emits `defeated` when `curr` is null. The `terminal_escape` logic in `computeBanditAnimPhase` + `renderBattlePhase` is unreachable. Either add a real discriminator (e.g. `prev.attackAttemptsMade >= 6`) or delete the variant and file a tracked follow-up for when backend exposes it. |
| M2 | 2/4 (Codex 5.4, Codex 5.5) | Pause handling: `updateBanditAnimation` freezes `walkFrameIdx=0` on pause, but `renderBattlePhase` still uses `Date.now()` for circle orbit, death flicker, win-cluster jitter. Paused worlds visibly animate. Fix: pass a frozen timestamp into `renderBattlePhase` or gate all `Date.now()` motion on `!worldPaused`. |
| M3 | Gemini unique | Missing `targetClanId` in `won` outcome type: `renderBattlePhase` has no reference to the *defeated* base — it uses `toAnchor` (destination) instead, so the march/whirlwind executes at the wrong location. Fix: add `targetClanId` to the `won` variant and populate from `prev.projectedTargetClanId` in the diff useEffect. |
| M4 | 2/4 (Codex 5.4, Codex 5.5) | `combatDim`/`combatFlash` shared layers not cleaned up on early returns (hidden, DEMO start, asset failure). Stale dim/flash alpha can linger across the next render. Fix: normalize alpha in the top-of-frame reset or on every early-return path. |
| M5 | Opus 4.7 unique | `BanditState.Defeated` (enum=4) not handled in `computeBanditAnimPhase`. If snapshot transiently shows `Defeated` before null, function falls through to `hidden` with no death sequence. Fix: add a `Defeated` branch that synthesizes a `defeated` outcome from bandit row data. |
| M6 | Gemini unique | Pixi sub-container creates Z-sort isolation boundary. `walker.zIndex` sorts bandits only within the container, not against bases in `worldDynamic`. Bandits can render on top of bases they should walk behind. Fix: add sprites directly to `worldDynamic` or dynamically update `container.zIndex` each frame to the walkers' max Y-value. |
| M7 | Gemini unique | Asset fallback gap: if walk sprites 404 but camp sprite loaded, `loadBanditAnimAssets` returns early (empty array check), leaving `assetsReady=false`. User sees no bandit at all. Fix: remove early return; set `assetsReady=true` if at least one texture loaded. |
| M8 | Opus 4.7 unique | Bandit is un-clickable in live mode: `redrawBandit` hides legacy `banditIcon` (the only pointertap handler) for non-DEMO. New walker/camp sprites have no event wiring. Fix: add `eventMode='static'` + `pointertap` to `animContainer.campSprite`. |
| M9 | Opus 4.7 unique | No unit tests for the exported `computeBanditAnimPhase` pure function. The planning doc explicitly lists 5 test scenarios. The M1/M3/H3 issues would have been caught by tests. At minimum: add vitest cases for Camped/ticksInState boundary, defeated battleT clamp, won battleT>1 hidden, and stale-outcome guard. |

---

## DEFER (LOW)

- walker alpha not reset between phases — fade leaks from terminal_escape (when wired up) into subsequent phases (Codex 5.5 L1, Opus 4.7 M4 as defensive gap). Reset `alpha=1` in top-of-frame reset loop.
- Partial sprite load: `pickWalkDirection` can choose empty texture array if only one direction loaded (Codex 5.5 L2). Explicit fallback to non-empty direction.
- camp_idle → camp_telegraph is a hard cut, not a morph (Opus 4.7 L1). Cross-fade alpha.
- Slight glow alpha discontinuity at 2.83 tick boundary (Opus 4.7 L2).
- Circle orbit always uses NE textures regardless of direction (Opus 4.7 L3).
- `BATTLE_T_FADE_END` and `BATTLE_T_WIN_WALK_END` both = 25.5/60 — comment or alias (Opus 4.7 L4).
- `WORLD_WIDTH + 50` terminal exit math fragile under future viewport rescaling (Opus 4.7 L5).
- `redrawBandit` called repeatedly in live mode to do nothing (Opus 4.7 L6).
- `lastBanditOutcomeRef` / `prevBanditRef` cleanup coupled to Pixi lifecycle, not own useEffect (Opus 4.7 L7).

---

- `won-vs-no_battle_advance` heuristic on `prev.projectedTargetClanId > 0`: Codex 5.4/5.5 flagged as MED concern. Opus 4.7 assessed as acceptable v1 fallback with a follow-up issue. Given forward-compat with PR #151 (backend no-target-advance), this is intentional design — document and file a follow-up. **Skip for this round.**
- DEMO_MODE backward compat: Gemini verified intact. Codex 5.5 + Opus 4.7 both confirmed. **Not a finding.**
- Memory/cleanup (`loadBanditAnimAssets` cancellation check, Pixi destroy cleanup): Opus 4.7 assessed as correct. **Not a finding.**
- Performance: `currentTickFloat` interpolation and Date.now() decoupling assessed as good pattern by Gemini. Pre-allocated walker/death sprites prevent GC. **Not a finding.**

---

## Per-model verdicts

- **Codex 5.4 ✓** — NEEDS_FIXES. 2 HIGH (cold-start T4 recovery, Camped-only guard failure), 3 MED, 1 LOW. Strong signal on seam bugs.
- **Codex 5.5 ✓** — NEEDS_FIXES. 2 HIGH (phase machine priority suppresses won/no_battle_advance, cold-start reload), 3 MED, 2 LOW. Best articulation of the priority inversion bug.
- **Opus 4.6 ✗** — Pending (0-byte at synthesis time; task b7trbqbrx still in flight).
- **Opus 4.7 ✓** — NEEDS_FIXES. 1 HIGH (legacy vignette preempts — unique and most critical finding), 6 MED, 7 LOW. Most comprehensive review; caught the battle-path-dead-code finding others missed.
- **Gemini 3.1 Pro ✓** — NEEDS_FIXES. 2 HIGH (phase priority + missing targetClanId in won), 3 MED, 2 LOW. Unique find: missing targetClanId makes battle execute at wrong map location.

---

## Cross-model overlap stats

- **3/4 consensus HIGH**: H2 (computeBanditAnimPhase priority inversion — won/no_battle_advance suppressed)
- **2/4 HIGH + 1/4 MED**: H3 (cold-start T4 reload loses outcome)
- **1/4 HIGH (Opus 4.7 unique)**: H1 (legacy vignette preempts — but uniquely impactful; makes entire battle path unreachable)
- **1/4 HIGH + 1/4 MED**: H4 (Camped-only guard misses Attacking/Defeated transitions)
- **4/4 MED/LOW overlap**: M1 (terminal_escape dead code) — every reviewer flagged it
- **2/4 MED**: M2 (pause Date.now() not frozen), M4 (shared layer dirty on early return)
- **1/4 unique MED**: M3 (Gemini: missing targetClanId), M5 (Opus 4.7: Defeated enum unhandled), M6 (Gemini: Z-sort isolation), M7 (Gemini: asset fallback gap), M8 (Opus 4.7: un-clickable bandit), M9 (Opus 4.7: no tests)

---

## Fix-round scope

**Must address before merge:**
1. Gate legacy `BanditAttackResolved` chain-event useEffect (`WorldMap.tsx:3840`) behind `!DEMO_MODE` OR invert container-yield in live mode (H1) — ~5 min
2. Invert phase priority in `computeBanditAnimPhase`: check `lastOutcome` time-window first, before live bandit steady state (H2) — ~30 min
3. Widen diff useEffect guard to handle `prev.state === Attacking` (H4) — ~20 min
4. Acknowledge cold-start T4 as known gap OR add backend `lastResolution` field (H3) — ~10 min if doc-only
5. Remove or discriminate `terminal_escape` variant (M1) — ~20 min
6. Fix `targetClanId` in `won` outcome type + renderBattlePhase routing (M3) — ~30 min
7. Handle `BanditState.Defeated` in `computeBanditAnimPhase` (M5) — ~15 min

**Good to include (recommend filing issues if not inlined):**
- M8 bandit un-clickable in live mode
- M2 pause Date.now() freeze
- M9 unit tests for computeBanditAnimPhase

**Total fix-round budget: ~2.5h codex**

---

## Decision

**Path A — R1 fix-round.** Dispatch codex on the 7 MUST items + M8 (clickability). M2, M4, M6, M7, M9 as filed issues.

**Liam decision needed:** H1 architecture choice — (a) gate legacy vignette useEffect behind DEMO_MODE entirely, or (b) add a `phaseMachineEnabled` flag so both systems coexist until the new one is validated. Option (a) is smaller and matches PR intent; option (b) is safer for production rollback. Liam call.

claude-do:
# Phase Super-Swarm Synthesis — PR #150 rev2 (head d8fcb1a)

**Models run:** Codex 5.4 ✓ | Codex 5.5 ✓ | Gemini 3.1 Pro ✓ | Opus 4.7 ✓ | Opus 4.6 ✗ (0-byte output after 40+ min — suspected model hang)
**PR:** feat(web): bandit attack animation — telegraph + battle + outcome variants
**Branch:** feat/bandit-animation-impl
**Diff size:** 991 lines
**Rev 2 context:** Fix-round on d8fcb1a — H1/H2/H4/M1/M3/M5/M8/S1 claimed resolved

## Summary

**Verdict: NEEDS_FIXES** — 4/4 completed reviewers agree. Two MUST FIX items before merge: a new HIGH severity zIndex regression that makes live-mode bandit visuals render *under* every base/clansman during battles (caught only by Opus 4.7), and a MEDIUM Defeated→null animation replay bug that restarts the death sequence from zero (3/3 other reviewers caught this independently). S1 pause freeze is BROKEN end-to-end (worldPaused never delivered by snapshot schema — 3/3).

H1, H2, and M1 are cleanly VERIFIED by all reviewers. The fix-round correctly isolated DEMO mode but introduced a container zIndex oversight and a diff-guard regression.

|---|---|---|---|---|
| **NEW-H1** | WorldMap.tsx:2612-2615 | Opus 4.7 only (1/4) | **HIGH** | **Bandit container zIndex 0 — live battles invisible under bases.** The new `banditAnimRef.container` is appended to `layers.worldDynamic` which has `sortableChildren = true`. The container never sets its own `zIndex`, so it defaults to 0 and gets z-sorted to the bottom of worldDynamic below every base, clansman, and travel dot (which all set `zIndex = Math.round(y)` = hundreds-to-thousands). During battle, bandits march toward and circle a base — the exact overlap case. The walker/death sprites set their own `zIndex` correctly, but those values only sort within the container; the container itself is at the bottom. The legacy `drawn.banditSprite` avoided this by adding directly to worldDynamic as a sibling (not a child of a container). **Fix (option A):** set `container.zIndex` each frame in `updateBanditAnimation` to `Math.max(walker.zIndex for visible walkers)` so the container floats above. **Fix (option B, preferred):** don't use a wrapper container — add campSprite/walkers/deathSprites/campGlow directly to `worldDynamic` as siblings, matching the legacy approach. |
| **M5-REG** | WorldMap.tsx:1235, ~1241-1250 | 3/3 (Codex 5.4+5.5, Gemini HIGH; Opus 4.7 MED) | **MEDIUM** | **Defeated→null replays death animation from removal tick.** When bandit is in `Defeated` state, the live `BanditState.Defeated` branch in `computeBanditAnimPhase` already synthesizes the death animation using `resolvedTick: bandit.stateEnteredTick`. But on the next snapshot tick when the backend purges the Defeated row (`curr=null`), `wasResolutionCandidate` is true (because `prev.state === Defeated`) and overwrites `lastBanditOutcomeRef` with `resolvedTick: tick` (the purge tick) — resetting `battleT` to 0 and replaying the entire 17s march/circle/death sequence. **Fix:** In the `if (!curr)` block, use `prev.stateEnteredTick` when `prev.state === BanditState.Defeated`: `resolvedTick: prev.state === BanditState.Defeated ? prev.stateEnteredTick : tick`. Or simpler: skip writing `lastBanditOutcomeRef` entirely when `prev.state === Defeated` — the live-synthesized animation already ran; once the row is gone and no outcome ref exists, the phase machine correctly returns `hidden`. |

## SHOULD FIX (MEDIUM)

| # | Models | Finding |
|---|---|---|
| **S1-BROKEN** | Codex 5.4 + Codex 5.5 + Opus 4.7 (3/4) | **worldPaused never delivered by snapshot contract.** `updateBanditAnimation()` checks `snap.worldPaused === true` and freezes `renderNow` — but `getSnapshot()` never returns `worldPaused` and the `worldSnapshot` schema doesn't define it. Field is always `undefined`; freeze branch never runs; `currentTickFloat()` still advances via `Date.now()` while world is paused. A paused battle can finish and hide before the player can see it. **Fix options:** (a) plumb `worldPaused` through schema + `getSnapshot()`, (b) make `currentTickFloat()` clamp when paused, (c) accept as known gap with TODO comment if `worldPaused` doesn't exist in the backend yet. |
| **M3-PARTIAL** | Codex 5.4 + Codex 5.5 + Opus 4.7 (3/4) | **Deleted-clan falls back to wrong surviving base.** `targetClanId` is correctly preserved in `won` outcomes. But if the target clan no longer exists in `drawn.bases`, the cascade falls back to `drawn.bases.find(b => b.clan.homeRegion === phase.regionKey)` — the first surviving base in the region, which may be an unrelated clan. Animation plays on top of the wrong base. No crash. **Fix:** use the region anchor offset directly (`fromAnchor + { x:40, y:30 }`) when exact base lookup fails, not another clan's base. |
| **M-2 (new)** | Opus 4.7 only (1/4) | **`lastBanditOutcomeRef` never cleared after animation window closes.** Persists for lifetime of WorldMap mount. If a new bandit spawns within 25.5s of a previous outcome, the stale priority window blocks the new bandit's `camp_idle` phase. If `currentTickFloat()` ever rewinds (snapshot anomaly), stale outcome could re-fire. **Fix:** After `battleT >= BATTLE_T_FADE_END` or on new bandit at different region, null the ref. Easy place: in the diff `useEffect`, after updating `prevBanditRef.current = curr`, check if existing outcome is more than 0.5 ticks old and null it. |

## DEFER (LOW)

- **M8 PARTIAL — death sprites not clickable** (Codex 5.4 LOW, Gemini PARTIAL, Opus 4.7 VERIFIED-acceptable): `deathSprites` intentionally excluded from `attachBanditPointer`. Fading tombstones are not meaningful interaction targets. Defer. File GH issue for completion tracking.

- **Unreachable post-walk camp branch** (Gemini MED, Opus 4.7 LOW): `BATTLE_T_WIN_WALK_END === BATTLE_T_FADE_END = 25.5/60`. The `else if (toAnchor)` hold-camp block inside `renderBattlePhase` won-path is unreachable — phase machine returns `hidden` at exactly the same boundary. Causes one-frame vanish before `camp_idle` takes over. Fix: reduce `BATTLE_T_WIN_WALK_END` to `22/60` for 3.5s overlap, or split `BATTLE_T_FADE_END` into defeat/win variants. File cleanup issue.

- **Dead `if (!DEMO_MODE)` branch in `redrawBandit`** (Gemini LOW, Opus 4.7 LOW): `drawn.banditIcon` is never set in live mode; `if (!icon || !text) return` exits first. The `!DEMO_MODE` block is unreachable. Remove or convert to assert comment.

- **Combat ticker still registered in live mode** (Opus 4.7 LOW): `combatCb` added per-frame despite `updateCombatVignette` early-returning for non-DEMO. Harmless waste. Defer.

- **Camped→null false defeat outcome** (Opus 4.7 LOW): If a bandit is removed while Camped (debug command or future no-target-despawn rule), diff produces a `defeated` animation for a bandit that never fought. Acknowledged in code comments as known approximation. File follow-up issue.

- **Plan doc placeholder** (Opus 4.7 LOW): `docs/planning/bandit-animation-impl-plan.md` is 6 lines referencing sections that don't exist in the doc. Flesh out or remove inline `§` references.

- **M5 not exercised end-to-end** (Codex 5.4 + 5.5 note): `getSnapshot()` currently strips `BanditState.Defeated`, so `prev=Attacking + curr=Defeated` never occurs from live data. Code is correct in isolation but untested by live traffic. Document as known gap for when backend filter is removed.

- **M3 combatHighlight z-order intent** (Opus 4.7 MED): Concern that future selection rings in `combatHighlight` won't coordinate with live bandit visuals. Correct-but-deferred observation — current PR has no such feature and the code comment explicitly defers `worldBanditCombat` layer to multi-bandit landing. Skip for this PR.

- **S1 verified by Opus 4.7 / BROKEN by Codex x2**: Opus 4.7 marks S1 VERIFIED (the `renderNow` freeze is present and correct for sprite timing). Codex 5.4 + 5.5 mark it BROKEN (the `worldPaused` field isn't plumbed through the schema, so the freeze never activates). Both are correct — the renderer-side code is correct but the data contract is missing. Classified as SHOULD FIX (schema gap), not a renderer bug.

## Per-model verdicts

| Reviewer | Verdict | HIGH | MED | LOW | Status |
|---|---|---|---|---|---|
| Codex 5.4 | NEEDS_FIXES | 0 | 2 | 1 | ✓ |
| Codex 5.5 | NEEDS_FIXES | 0 | 3 | 0 | ✓ |
| Gemini 3.1 Pro | NEEDS_FIXES | 1 (M5-REG) | 1 | 1 | ✓ |
| Opus 4.7 | NEEDS_FIXES | 1 (zIndex) | 3 | 5 | ✓ |
| Opus 4.6 | — | — | — | — | ✗ (0-byte after 40+ min) |

## Cross-model overlap stats

```
NEW-H1 (container zIndex = 0 → invisible under bases) — 1/4 (Opus 4.7 HIGH; missed by all others)
M5-REG (Defeated→null replays death)                  — 4/4 (Gemini HIGH, Codex 5.4+5.5+Opus 4.7 MED)
S1-BROKEN (worldPaused not in schema)                 — 3/4 (Codex 5.4+5.5+Opus 4.7 — Gemini marked VERIFIED)
M3-PARTIAL (deleted-clan wrong base)                   — 3/4 (Codex 5.4+5.5+Opus 4.7 MED — Gemini VERIFIED)
M-2 (outcome ref never cleared)                        — 1/4 (Opus 4.7 MED)
H1 VERIFIED (no orphan live handlers)                  — 4/4 VERIFIED
H2 VERIFIED (stale outcome falls through)              — 4/4 VERIFIED
M1 VERIFIED (terminal_escape gone)                     — 4/4 VERIFIED
M8 PARTIAL (death sprites excluded from pointer)       — 4/4 (all rate as acceptable/defer)
```

## Fix verification checklist (consensus across 4 reviewers)

| Fix | Status | Confidence |
|-----|--------|-----------|
| H1 — animContainer sole live visual, no orphan handlers | VERIFIED | 4/4 |
| H2 — lastOutcome priority, stale falls through | VERIFIED | 4/4 |
| H4 — Attacking/Defeated in diff guard | PARTIAL (Defeated removal creates replay) | 4/4 |
| M1 — terminal_escape removed | VERIFIED | 4/4 |
| M3 — targetClanId in won outcome | PARTIAL (deleted clan wrong fallback) | 3/4 |
| M5 — BanditState.Defeated handled | PARTIAL (Defeated→null replay new bug) | 4/4 |
| M8 — sprite clickability | PARTIAL/acceptable (death sprites excluded) | 4/4 |
| S1 — pause-frozen timestamp | BROKEN (worldPaused not in snapshot schema) | 3/4 |

## Decision

**Recommend Path A R3 fix-round** (~20-30 min codex, surgical, all in WorldMap.tsx).

Scope:
1. **NEW-H1 (MUST):** Fix container zIndex. Option B (preferred per Opus 4.7): remove the wrapper Container and add campSprite/walkers/deathSprites/campGlow directly to `worldDynamic` as siblings, matching legacy approach. Each sprite already sets its own `zIndex = Math.round(y)` per frame. Or option A: update `container.zIndex` each frame in `updateBanditAnimation` to the max visible child zIndex.
2. **M5-REG (MUST):** In the `if (!curr)` block of the snapshot diff `useEffect`, when `prev.state === BanditState.Defeated`, skip writing `lastBanditOutcomeRef` entirely (the live-synthesized animation already ran; bandit going hidden is correct after Defeated→null).
3. **S1 fix (SHOULD):** If `worldPaused` exists in the backend schema, plumb through `getSnapshot()`. If not, add a TODO comment marking the freeze as pending backend field and accept as known gap. Do not leave broken code that silently does nothing.
4. **M3 fix (SHOULD):** In `renderBattlePhase` cascade, when `targetCenter` lookup fails, fall back to region anchor offset rather than first surviving base in region.
5. **M-2 (SHOULD):** Clear `lastBanditOutcomeRef.current = null` in the diff `useEffect` when the existing outcome's window has expired (`currentTickFloat - resolvedTick >= BATTLE_T_FADE_END`).

Total fix-round budget: ~20-30 min codex.

**Liam decision needed:** Whether S1 requires backend schema work before merge or can be deferred with TODO comment. NEW-H1 (zIndex) is a clear MUST FIX — bandits are invisible during battles in live mode.

claude-do:
# Phase Super-Swarm Synthesis — PR #150 (head 23c04e4)

**Models run:** Codex 5.4 ✗ (prompt-echo, no review content) | Codex 5.5 ✓ | Opus 4.6 ⧖ (in flight at synthesis time — stdin dispatch) | Opus 4.7 ✓ | Gemini 3.1 Pro ✓
**Branch:** feat/bandit-animation-impl
**Diff size:** 994 lines
**Reason:** R3 fix-round verification — wrapper-container removal, M5-REG, M-2, M3-PARTIAL, S1

---

## Summary

**Verdict: NEEDS_FIXES (minor) — no HIGHs, 3 confirmed MEDs needing patches before merge.**

R3 fixes are correctly applied across all reviewers: wrapper-container removal is clean, sprites attach directly to `worldDynamic`, zIndex uses `Math.round(y)` consistently across all sprite types, M5-REG (Defeated→null no longer triggers death replay) is confirmed working, and M-2 stale ref cleanup is sound. The phase machine + snapshot-diff architecture is well-factored.

Three medium-severity issues remain: an async re-entrancy race in `loadBanditAnimAssets` that can double-create sprites (2/3 models), a snapshot-diff identity bug where `prev.state === Defeated + curr at different region` fires a spurious 'won' animation (2/3 models), and an unscaled neutral-region anchor offset (1/3 models, independently found by orch analysis). All three are cheap to fix (~10 lines total).

---

## MUST FIX — Multi-model consensus

|---|---|---|---|---|
| MF1 | `WorldMap.tsx`, `loadBanditAnimAssets` | 2/3 (Gemini HIGH, Opus 4.7 MED) | HIGH consensus | **Async re-entrancy race → double sprite add and permanent sprite leak on `worldDynamic`.** `assetsReady` flag is only set after `Promise.all` resolves. Two concurrent calls to `createBanditVisuals` (HMR, StrictMode double-invoke, snapshot reconnect) both pass `!assetsReady`, both append walkers/death/camp/glow to `worldDynamic`, and the second pass overwrites `banditAnimRef.current.walkers` so the first batch is orphaned as invisible persistent nodes. Fix: add an `isLoading` boolean sentinel before the await; set it true before `Promise.all`, reset in `finally`. |
| MF2 | `WorldMap.tsx`, snapshot-diff `useEffect` | 2/3 (Codex 5.5 MED, Opus 4.7 MED) | MED consensus | **`prev.state === Defeated + curr at different region` fires spurious 'won' outcome.** `wasResolutionCandidate` includes `BanditState.Defeated`. M5-REG fix only skips the `!curr` arm. If backend delivers `Defeated → Spawned/Camped at different region` in one snapshot tick (purge + respawn collapsed), the `curr.region !== prev.region` arm fires and writes a 'won' outcome with the old bandit's `projectedTargetClanId` — march→circle→flash→walk animation plays for a dead bandit. Fix: guard the second arm with `&& prev.state !== BanditState.Defeated`. |
| MF3 | `WorldMap.tsx`, `renderBattlePhase` ~line 734 | 1/3 (Opus 4.7 MED) + orch analysis | MED | **Neutral-region fallback offset `{x: +40, y: +30}` not scaled by `layoutScale`.** Every other offset in this function multiplies by `layoutScale`. At mobile viewport scales the battle pivot will be nearly at camp; at large viewports it's barely offset. Fix: `{ x: fromAnchor.x + 40 * layoutScale, y: fromAnchor.y + 30 * layoutScale }`. |

---

## SHOULD FIX (MEDIUM)

| # | Models | Finding |
|---|---|---|
| S1 | Codex 5.5 only | **`resolvedTick` stamped at current snapshot tick, not actual resolution tick.** If snapshot arrives delayed (prev.nextActionTick <= tick by >1 tick), animation restarts late. Fix: seed from `prev.nextActionTick` for Camped→resolved, `prev.stateEnteredTick` for Defeated live synthesis. Deferrable if delayed snapshots don't occur in practice. |
| S2 | Gemini MED | **Z-fighting from jitter: telegraph/cluster walkers can clip below camp glow.** Glow zIndex = `Math.round(anchor.y) - 1`; walkers get `Math.round(walker.y)` after 1.5px jitter. If anchor.y decimal is small, walker rounds below glow. Fix: compute walker zIndex from baseline `anchor.y` before jitter, not from post-jitter `walker.y`. |

---

## DEFER (LOW)

- **L1 (Opus 4.7):** Cold-reload mid-T4 cuts off death animation — acknowledged as known limitation in `docs/planning/bandit-animation-impl-plan.md`. Needs backend `lastResolution` field. Not blocking.
- **L2 (Opus 4.7):** Tier-3 target base lookup (`drawn.bases.find(b => b.clan.id === targetClanId)`) is region-agnostic — could match a base in a different region. Add sanity radius or drop tier-3.
- **L3 (Opus 4.7):** Orbit walkers always use NE texture set (hardcoded `y: -1` in tangent calc). SE texture unused for bottom half of orbit. Minor visual.
- **L4 (Opus 4.7):** `BATTLE_T_WIN_WALK_END === BATTLE_T_FADE_END` — the "glow camp at destination" `else if` in the won path can never fire because `computeBanditAnimPhase` returns `hidden` at `battleT >= BATTLE_T_FADE_END`. Dead branch. Either drop it or set `BATTLE_T_WIN_WALK_END = 22/60` to create the gap the comment implies.
- **L5 (Opus 4.7):** Walker zIndex ties at same rounded y in cluster/circle phases — Pixi falls back to insertion order. Add `+ i` to deterministically order siblings.
- **L6 (Gemini LOW):** Per-frame `drawn.bases.find()` on 60fps ticker for base target lookup. Negligible on small maps. Cache at battle-phase start if needed.
- **L7 (Codex 5.5, as MED — re-classified):** `resolvedTick` vs `prev.nextActionTick` timing drift on delayed snapshots — moved to S1 SHOULD FIX above.

---

- Gemini noted cold-reload mid-defeat "successfully latches onto live state and animates the remainder" under M5-REG check — this is correct per the existing `computeBanditAnimPhase` Defeated branch, not a new issue.
- Codex 5.5 flagged a conflict between the PR test-plan claim "Reload during mid-T4: animation phase resumes correctly" and the `getSnapshot.ts` filtering of Defeated→null. This is a doc/test-plan accuracy issue, not a code bug. Recommend narrowing the test-plan claim to "non-reload live path" in a follow-up PR comment, not a code change.

---

## R3 Verification Results (per checklist)

| Check | Result |
|---|---|
| A) Wrapper-container removal — sprite cleanup/destroy | CLEAN (3/3 reviewers). Direct-to-worldDynamic attachment follows existing pattern. No new traversal assumptions broken. Pixi `app.destroy()` cascades cleanup correctly. |
| A2) Sprite leak if cleanup before init | FLAGGED — MF1. Race window between first call and `assetsReady=true`. |
| B) zIndex correctly computed for all sprite types | CLEAN — camp, walkers, death, glow all get `Math.round(y)` (or `y-1` for glow) on every frame. Glow jitter edge case flagged in S2 but not cross-confirmed as HIGH. |
| C) M5-REG: Defeated→null no longer triggers death replay | CONFIRMED FIXED by 3/3 reviewers. The `prev.state !== BanditState.Defeated` guard in the `!curr` arm is correct. |
| C2) Cold-reload mid-defeat (Defeated→null, stale ref) | Correctly degrades to `hidden` (no crash). Death animation replays from live snapshot Defeated state. Known cold-start gap documented. |
| D) M-2 stale ref cleanup race | CLEAN (Gemini explicit). BATTLE_T_FADE_END evaluation correctly uses `currentTickFloat()` against `resolvedTick`, not snapshot tick. No race with same-tick new spawn. |
| E) M3-PARTIAL ALIVE clan target lookup precedence | CLEAN (Gemini: "extremely robust"). Three-pass: clan in region (string) → clan in region (numeric) → global clan → neutral anchor. No inversion risk found. |
| F) New issues from R3 wrapper removal | MF1 (async race). MF2 (Defeated→won spurious outcome, pre-existing edge case now more reachable). MF3 (unscaled offset). |

---

## Per-model verdicts

- **Codex 5.5** ✓ — NEEDS_FIXES, 0 HIGH, 3 MED, 0 LOW. Focused on snapshot-diff identity correctness and resolvedTick timing. Did not flag the async race.
- **Gemini 3.1 Pro** ✓ — NEEDS_FIXES, 1 HIGH (async race = MF1), 2 MED (jitter z-fighting + base find perf), 0 LOW. Strong R3-specific verification, found the most critical issue.
- **Opus 4.7** ✓ — NEEDS_FIXES (minor), 0 HIGH, 3 MED, 8 LOW. Most comprehensive review. Caught MF2, MF3, L4 (dead branch), L3 (orbit texture), and confirmed all R3 fixes clean. Recommended fix scope: M1+M2+M3 ≈10 lines.
- **Codex 5.4** ✗ — prompt-echo failure, no review content produced.
- **Opus 4.6** ⧖ — dispatch in flight at synthesis time (stdin re-dispatch after arg-size failure). Will update if output arrives with material new findings.

---

## Cross-model overlap stats

- MF1 (async race): 2/3 ✓ reviewers flagged (Gemini HIGH, Opus 4.7 MED)
- MF2 (Defeated→won spurious): 2/3 ✓ reviewers flagged (Codex 5.5 + Opus 4.7 MED)
- MF3 (unscaled neutral offset): 1/3 ✓ (Opus 4.7 only)
- S2 (jitter z-fighting): 1/3 ✓ (Gemini only)
- R3 fixes confirmed clean: 3/3 ✓ (wrapper removal, M5-REG, M-2, M3-PARTIAL)

---

## Decision

**Recommend Path A R4 fix-round** — 3 targeted patches, ~10 lines, no test-surface change.

Scope:
1. `loadBanditAnimAssets`: add `isLoading` sentinel before `Promise.all` await; reset in `finally`. (MF1)
2. Snapshot-diff `useEffect`: add `&& prev.state !== BanditState.Defeated` guard to the `curr.region !== prev.region` branch. (MF2)
3. `renderBattlePhase` neutral-region fallback: multiply `+40` and `+30` offsets by `layoutScale`. (MF3)

Optional in same round (SHOULD FIX):
4. S2 — telegraph/cluster walker zIndex: compute from `anchor.y` before jitter.

Total fix-round budget: ~15 min codex.

**No Liam decision needed** — all findings are code quality / defensive hardening, no scope change or architectural shift.

claude-do:
**[SYNTHESIS UPDATE — Opus 4.6 result now included]**

Opus 4.6 completed after initial synthesis post. Verdict: **CLEAN** (no MED/HIGH). Key additions:
- Rates MF1 async race as LOW (single callsite, currently unreachable) — disagrees with Gemini HIGH / Opus 4.7 MED, but the fix is still recommended as defensive hardening.
- All R3 checks confirmed clean (unanimous 4/4 reviewers).
- New unique LOWs: DEMO_MODE loads unused sprites (L10), `terminal_escape` not yet in type system (L9, documented follow-up), cluster comment timestamp wrong (L8).

**Overall verdict unchanged: NEEDS_FIXES (minor) — MF1+MF2+MF3 patches before merge.**

Full synthesis updated at: `docs/reviews/pr150-rev3-synthesis.md`

claude-do:
# Phase Super-Swarm Synthesis — PR #150 rev 4 (head 7115264)

**Models run:** Codex 5.4 ✓ | Codex 5.5 ✓ | Opus 4.6 ✓ | Opus 4.7 ✓ | Gemini 3.1 Pro ✓ | Codex 5.3 ✗ (unavailable — not in lineup per canonical config)
**Phase:** feat/bandit-animation-impl
**Diff size:** 1002 lines
**Reason:** Rev 4 super-swarm after R4 fix-round (commit 7115264)

---

## Summary

**Verdict: CLEAN (with one pre-existing LOW queued for follow-up)**

All 4 R4 fixes (MF1 isLoading sentinel, MF2 Defeated→different-region guard, MF3 layoutScale neutral fallback, S2 walker zIndex baseline) verified PASS by all 5 reviewers. The one contested finding — an unscaled `-30` Y offset on the primary `targetBase` branch (`WorldMap.tsx:2861`) — was unanimously identified but is pre-existing code from before this PR. Gemini graded it HIGH; Codex 5.4 graded it MEDIUM; Opus 4.6 + 4.7 graded it LOW/pre-existing; Codex 5.5 did not flag it. It was NOT introduced by the R4 fix-round and is NOT a regression. **Recommend merge.** File a follow-up issue for the pre-existing offset.

---

None. No merge-blocking findings.

---

## SHOULD FIX (MEDIUM)

None identified as introduced by this PR.

---

## DEFER (LOW) — pre-existing, follow-up issues to file

| # | File:line | Models | Finding |
|---|---|---|---|
| L1 | WorldMap.tsx:2861 | 4/5 | Unscaled `targetBase.container.y - 30` Y offset. MF3 fix correctly addressed the neutral fallback branch; this adjacent offset on the real-target branch was missed. **Pre-existing** — not introduced by R4. At normal layoutScale (~1.0) purely cosmetic; will drift at extreme zoom levels. One-line fix: `y: targetBase.container.y - 30 * layoutScale`. |
| L2 | WorldMap.tsx:2569-2572 | 1/5 (codex-5-4) | `loadBanditAnimAssets` still called in DEMO_MODE — allocates textures never shown. Cosmetic waste, pre-existing design. |
| L3 | WorldMap.tsx:3003-3019 | 1/5 (opus-4-7) | Dead-code `else if (toAnchor)` branch after win-walk transit — unreachable by construction. Pre-existing. Delete in cleanup pass. |
| L4 | WorldMap.tsx:2653-2659 | 1/5 (opus-4-7) | Death/tombstone sprites lack `attachBanditPointer`. Likely intentional design ("dead bandits not selectable"). Confirm with Liam. |
| L5 | WorldMap.tsx:2744 | 2/5 (codex-5-4, opus-4-7) | `(snap as any)?.worldPaused` cast — dormant TODO path, well-commented, not hiding an active bug. |

---

- **Gemini HIGH on L1 (unscaled `-30`)**: Severity over-escalated. This is identical code that existed before this PR; it was not introduced or worsened by the R4 fix-round. The MF3 scope was to fix the neutral fallback specifically (which it did). Downgraded to LOW/follow-up. Four other reviewers confirmed pre-existing.

---

## Fix verification checklist (5/5 consensus)

| Fix | Codex 5.4 | Codex 5.5 | Opus 4.6 | Opus 4.7 | Gemini | Verdict |
|---|---|---|---|---|---|---|
| MF1 isLoading sentinel | PASS | PASS | PASS | PASS | PASS | PASS 5/5 |
| MF2 Defeated→different-region guard | PASS | PASS | PASS | PASS | PASS | PASS 5/5 |
| MF3 layoutScale neutral fallback | PASS | PASS | PASS | PASS | FAIL* | PASS 4/5 |
| S2 walker zIndex baseline | PASS | PASS | PASS | PASS | PASS | PASS 5/5 |

*Gemini's MF3 FAIL is for the adjacent unscaled line (2861) that was not part of the MF3 fix scope. The specific fix (line 2867 neutral fallback) is PASS per all reviewers.

---

## Per-model verdicts

- **Codex 5.5**: CLEAN — all 4 fixes pass, no findings. Merge recommended.
- **Codex 5.4**: NEEDS_FIXES — calls out unscaled `-30` at line 2861 as MEDIUM (same-class bug). Also flags LOW on asset fallback comment accuracy. Merge conditional on addressing.
- **Opus 4.6**: CLEAN — 4 fixes pass. Flags unscaled `-30` as pre-existing LOW, `as any` as LOW. Merge recommended.
- **Opus 4.7**: CLEAN — 4 fixes pass. Most thorough fresh scan: found dead-code branch, pre-existing unscaled `-30`, death pointer UX nit. All pre-existing, all LOW. Merge recommended.
- **Gemini 3.1 Pro**: NEEDS_FIXES — escalates unscaled `-30` to HIGH. All 4 named fixes verified. Merge conditional on addressing.

---

## Cross-model overlap stats

- **4/5 models flagged** unscaled `targetBase.container.y - 30` (line 2861): Gemini HIGH, Codex 5.4 MED, Opus 4.6 LOW, Opus 4.7 LOW. Codex 5.5 did not flag. Strong signal on a real issue — but pre-existing scope question settles it at LOW.
- **5/5 models PASS** MF1 isLoading sentinel.
- **5/5 models PASS** MF2 Defeated guard.
- **5/5 models PASS** S2 zIndex baseline.
- **4/5 models PASS** MF3 (Gemini FAIL scoped to adjacent line, not the named fix target).
- **1/5** flagged DEMO_MODE texture waste (L2), dead code (L3), death pointers (L4).

---

## Decision

**Recommend: MERGE CLEAN.** All 4 R4 fixes are verified correct and complete by 5/5 reviewers. The one contested finding (unscaled `-30` at line 2861) is pre-existing code predating this PR, confirmed not introduced by the R4 fix-round.

**Follow-up issue to file post-merge:**
1. `WorldMap.tsx:2861` — Scale `targetBase.container.y - 30` with `layoutScale` (one-line fix, batch with any other pre-existing unscaled offsets in `renderBattlePhase`).
2. `WorldMap.tsx:3003-3019` — Remove dead `else if (toAnchor)` branch.
3. Confirm death-sprite pointer-interactivity is intentional design.

**Total follow-up budget:** ~10 min in a future cleanup pass. Not blocking merge.

**Liam decision needed:** None — swarm is unanimous that the R4 fixes are correct. Pre-existing L1 can ship and be addressed in a follow-up.

Issue #229, opened 2026-05-11
v2.4.0 Android: home screen JSON decode error + mint network mismatch (mainnet vs devnet)


claude-do:
**Two bugs reported 2026-05-11 against v2.4.0 Android build:**

## Bug A — Home screen JSON decode error

The Hearth home screen shows raw error text:

> Could not decode getSnapshot:getSnapshot response into so: {"status":"success","value":{"activeBanditId":1.0,"bandit":{"attackPower":60.0,"id":1.0,"nextActionTick":12.0,"projectedTargetClanId":0.0,"region":4.0,"state":1.0,"stateEnteredTick":11.0,"tier":3.0},"clans":[{"baseLevel":1.0,"baseRegion":1.0,"blueprintBalance":"0",...

The Convex server is sending integer-valued numbers as floats (\`1.0\`, \`60.0\`, etc.). The Android Kotlin parser (kotlinx.serialization with \`Int\`/\`Long\` field types) rejects \`1.0\` because it expects \`1\`.

**Likely root cause:** Convex JSON serialization preserves numeric types (Int → \`1\`, Float → \`1.0\`). Somewhere on the server side, an integer field is being coerced to a JS \`Number\` then re-serialized as \`1.0\`. OR — more likely — the Kotlin client uses a strict serializer that should be configured to accept either form.

**Fix candidates:**
- (a) Configure kotlinx.serialization with \`coerceInputValues = true\` AND/OR a custom \`Long\`/\`Int\` deserializer that accepts both \`1\` and \`1.0\` for integer fields.
- (b) Server-side: normalize integer fields to strict integers before responding. This is harder because Convex types are JS-native.

Recommended fix: (a) — make the Android parser permissive on integer-typed fields.

## Bug B — Network mismatch on mint devnet GOLD

Liam's wallet (chaintail.skr, default network mainnet) shows:

> Network mismatch — Your current network is set to mainnet, but this transaction is for devnet. Switch to the correct network before signing.

The Android app builds a devnet-targeting transaction; the wallet defaults to mainnet → mismatch dialog.

**Likely root cause:** The MWA session establishment doesn't request the \`devnet\` cluster, OR the transaction's RPC endpoint doesn't match the cluster identifier the wallet sees.

**Fix candidates:**
- (a) Pass the cluster identifier (\`solana:devnet\` per Solana Mobile MWA spec) when opening the MWA session in \`MwaClient.kt\`.
[...]

PR #231, opened 2026-05-11
fix(android): home screen JSON decode + devnet cluster routing (#229)

claude-do:
Closes #229

## Summary

Two v2.4.0 Android regressions:

**Bug A — Hearth home screen JSON decode error.** Convex serializes integer-typed fields (`bandit.id`, `bandit.tier`, `seasonStartTick`, etc.) as JSON floats (`1.0`, `60.0`), and kotlinx.serialization's default `Int`/`Long` decoders reject them. Result: home screen showed raw error text on every launch.

- New `ConvexJsonNumber.normalizeConvexNumbers()` walks the parsed `JsonElement` tree once before `decodeFromJsonElement` and rewrites tokens of literal shape `^-?\d+\.0+$` to `Long` via `substringBefore('.').toLongOrNull()`. **String-based on purpose** — going through `Double` would silently collapse 64-bit integers > 2^53.
- Out-of-Long-range tokens fall back to the original primitive so `Double`-typed fields still decode.
- Wired into `ClanWorldConvexClient.callConvex()` (the home/Hearth code path).
- Dropped now-incompatible `ClanSummary.baseLevel` / `baseRegion` / `wallLevel` / `monumentLevel` / `livingClansmen` (Convex emits these as bare numbers but the cockpit demo had typed them `String?` — they were unused).
- Dropped unused `WorldSnapshot.leaderboard` + `LeaderboardEntry` (UI builds rows from `snap.clans`).

**Bug B — Mint flow surfaced wallet "Network mismatch (mainnet vs devnet)".** Owner sign-in opened a `Solana.Mainnet` MWA session that primed the wallet on mainnet, then the devnet GOLD mint tripped the wallet's network-mismatch dialog.

- Introduced `ClanWorldMwaCluster = Solana.Devnet` as the single source of truth for every MWA session. Both `MwaClient.kt` (mint/sign-and-send) and `owner/Mwa.kt` (Ælder sign-in) now read from it.
- Added `MwaResult.WrongNetwork`. `classifyFailure()` detects narrow keyword set: `"network mismatch"`, `"wrong network"`, `"cluster mismatch"`, `"wrong cluster"`, `"invalid cluster"`, `"switch to the correct network"`, `"incorrect network"` — generic `"network request failed"` errors fall through to `Error`.
- Surfaced `WrongNetwork` in every call site (`ConnectViewModel`, `ClanWorldApp` mint flow, `SteeringConsole`, `ForgeScreen`, `StrategyEditor`, `HireModal`) with in-app guidance: "switch your wallet to Solana Devnet, then try again".

Local 3-tier swarm review (gemini-3-pro) ran two rounds; both pre-merge fix-rounds landed:
- r1 → r2: tightened `WrongNetwork` regex, switched normalizer to string-based parse, extracted `ClanWorldMwaCluster` constant.
- r2 → r3: fixed `normalizeLeavesOutOfLongRangeAlone` test to use an unquoted JSON number so the overflow guard actually fires.

22 unit tests, 0 failures. Debug APK assembles clean.

## Test plan
- [x] Mobile gradle build (`./gradlew :app:assembleDebug`) succeeds with map/terminal URLs set
- [x] Unit tests (`./gradlew :app:testDebugUnitTest`) — 22 pass, 0 fail
- [x] `ConvexJsonNumberTest` replays the v2.4.0 error payload through normalizer + WorldSnapshot decode
- [x] `MwaClusterTest` pins `ClanWorldMwaCluster == Solana.Devnet` (assertSame) so future Mainnet swap fails CI
- [ ] UAT: home screen loads without parse error on v2.4.1 build
- [ ] UAT: mint 100K GOLD without seeing wallet mainnet/devnet mismatch dialog (or sees clear in-app guidance if wallet refuses to switch)

### DIFF
--- a/apps/clan-world-mobile/app/src/main/java/world/clan/app/data/ClanWorldConvexClient.kt
+++ b/apps/clan-world-mobile/app/src/main/java/world/clan/app/data/ClanWorldConvexClient.kt
@@ -139,7 +139,8 @@ class ClanWorldConvexClient(private val convexUrl: String) {
       }
       val value = root["value"]
         ?: throw ConvexException("Convex $path returned no `value`: $body")
-      runCatching { json.decodeFromJsonElement<T>(value) }
+      val normalized = value.normalizeConvexNumbers()
+      runCatching { json.decodeFromJsonElement<T>(normalized) }
         .getOrElse {
           throw ConvexException(
             "Could not decode $path response into ${T::class.simpleName}: $body",
--- a/apps/clan-world-mobile/app/src/main/java/world/clan/app/data/ConvexJsonNumber.kt
+++ b/apps/clan-world-mobile/app/src/main/java/world/clan/app/data/ConvexJsonNumber.kt
@@ -4,7 +4,6 @@ import kotlinx.serialization.json.JsonArray
 import kotlinx.serialization.json.JsonElement
 import kotlinx.serialization.json.JsonObject
 import kotlinx.serialization.json.JsonPrimitive
-import kotlinx.serialization.json.doubleOrNull
 
 /**
  * Convex's JSON output renders ALL numeric fields as floats — e.g. an
@@ -17,6 +16,13 @@ import kotlinx.serialization.json.doubleOrNull
  * rewrites every whole-number Double primitive (e.g. `1.0` → `1`) so the
  * default `Long`/`Int` decoders accept it. Fractional values are passed
  * through untouched so `Double`-typed fields still decode correctly.
+ *
+ * The matcher is intentionally string-based — going through Double would
+ * silently collapse 64-bit integers > 2^53 (e.g. timestamps in
+ * nanoseconds) since `asLong.toDouble() == d` is still true after the
+ * binary64 precision floor. We only rewrite tokens of the literal shape
+ * `<digits>.0` (optionally signed), which is exactly the shape Convex
+ * emits for v.number()-typed integer columns.
  */
 internal fun JsonElement.normalizeConvexNumbers(): JsonElement = when (this) {
   is JsonObject -> JsonObject(mapValues { (_, v) -> v.normalizeConvexNumbers() })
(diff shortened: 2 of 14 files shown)


claude-do:
**Tier 1 (Claude subagent)** · round 1 (orch-swarm) · head `fdbe11f`

**Verdict: CLEAN** — no blocking issues.

### Bug A — ConvexJsonNumber normalizer

Regex `^-?\d+\.0+$` correctness checked across edge cases:

| case | match | result | ok |
|---|---|---|---|
| `0.0` / `-1.0` / `1.000000` | Y | → Long | ✓ |
| `1.5` / `0.25` | N | float-literal preserved | ✓ tested |
| `1e2` / `1.5e2` / `2E1` | N | preserved (test locks behavior) | ✓ |
| `-0.0` | Y | `-0` → 0L | ✓ tested |
| `99999999999999999999.0` | Y, but `toLongOrNull()` overflow → fallback to original | ✓ tested |
| strings / booleans / `JsonNull` | skipped via `isString` or non-matching content | ✓ tested |

Applied at the single `callConvex()` deserialization boundary; all three entry points (query/mutation/action) route through it. **No bypass.**

**Note:** there's a second, independent client at `data/convex/ConvexClient.kt` (cockpit UI) that does NOT use the normalizer. That's intentional and correct — its wire shapes (`SnapshotWire`, `ClanWire`, etc.) declare integer-looking fields as `Double`, which kotlinx already accepts from `1.0`. No regression risk.

**Dropped fields verified zero-consumer:** `ClanSummary.{baseLevel,baseRegion,wallLevel,monumentLevel,livingClansmen}` and `WorldSnapshot.leaderboard` — `grep -rn` confirms no references in `apps/clan-world-mobile/`. `HearthScreen.kt`'s `state.leaderboard` is a UI-state field built locally in `HearthViewModel.refresh()` from `ClanSummary` projections, NOT the dropped wire field. `ignoreUnknownKeys=true` drops the server-emitted fields silently. ✓

### Bug B — Devnet cluster routing

`grep -rn "Solana\.\(Mainnet\|Devnet\|Testnet\)" apps/clan-world-mobile/` returns only `MwaCluster.kt:19` (source of truth) + tests + kdoc reference. `grep -rn "mwa.blockchain\|.blockchain\s*="` returns only `Mwa.kt:48` and `MwaClient.kt:57` — both pointing at `ClanWorldMwaCluster`. **Single source of truth honored.**

**WrongNetwork classifier (7 keywords, case-insensitive via `.lowercase()`):**
- "network mismatch", "wrong network", "cluster mismatch", "wrong cluster", "invalid cluster", "switch to the correct network", "incorrect network"
- Narrow enough not to swallow generic transport errors ("network request failed" ≠ any keyword) ✓
- Covers Phantom's actual `"Network mismatch (mainnet vs devnet)"` per issue ✓
- Theoretical gap: localized-only wallet error strings would fall through to `MwaResult.Error`. Acceptable — strictly additive vs broken baseline, not a regression. Worth a follow-up issue if i18n-robust classification is wanted, NOT a blocker.

**6 UI consumers all handle `WrongNetwork`:** ClanWorldApp.kt:286, HireModal.kt:140, ForgeScreen.kt:120, SteeringConsoleScreen.kt:176, StrategyEditorScreen.kt:166, ConnectViewModel.kt:190 — verified with grep against the sealed-class branches.

### Tests

- 10 normalizer tests + 1 end-to-end `WorldSnapshot` decode using the actual v2.4.0 error payload from #229.
- 2 cluster tests (single source-of-truth + SDK rename guard).
- Coverage is solid for what's a small, targeted patch.

### Observations (non-blocking)

1. **Out-of-scope changes in PR**: also touches `apps/server/convex/*` (gold.ts reset-lock + idempotent-replay + clan-exists checks) and `packages/contracts/src/diamond/lib/LibBanditEvents.sol` (~+50 lines). Merge-from-dev artifacts based on commit history, but worth PM confirming these were intentional vs. reviewed under a sibling PR.
2. `Mwa.kt` `Solana` import correctly removed and replaced with `ClanWorldMwaCluster`.
3. `MwaClient.kt:37` stale kdoc ("V3 doesn't transact on Solana") was correctly updated to reflect that GOLD mint DOES broadcast.
4. Cosmetic: `MwaClusterTest.solanaMainnetIsNotDevnet` test name is slightly misleading — it asserts the SDK constant string, not inequality. Doesn't affect correctness.

**No HIGH/MED bug introduced. Ship.**

#claude-subagent

PR #232, opened 2026-05-11
[codex] Harden Elder runtime tooling

claude-do:
## Summary
- mirror local Elder runtime settings into `runtime/elders` and install a fail-closed Bash guard
- add `elder bulletin post` for public bulletin messages
- add `elder-mcp` stdio tools, including direct JSON `submit_orders` and `post_bulletin`

## Validation
- `pnpm --filter @clan-world/agents test`
- `pnpm --filter @clan-world/agents typecheck`
- hook smoke tests for allowed date/order/bulletin and denied cat/cross-clan view
- MCP stdio initialize/tools-list smoke

## Notes
No issue number was provided, so this PR does not include a `Closes #N` line.

PR #236, opened 2026-05-11
Dev

mcorrig4:


mcorrig4:
@codex please review

claude-do:
# Phase Super-Swarm Synthesis — PR #236 (head d77f338)

**Models run:** Codex 5.3 ✗ (prompt-echo, no review produced) | Codex 5.4 ✓ | Codex 5.5 ✓ | Opus 4.6 ✗ (0 bytes, 3 attempts, model unavailable) | Opus 4.7 ✓ | Gemini 3.1 Pro ✓
**Phase:** v2.4.1 dev→main release (PR #230 + #231 + #234 + #232)
**Diff size:** 1796 lines

---

## Summary

**Verdict: NEEDS_FIXES** — 4 active reviewers (Codex 5.4, Codex 5.5, Opus 4.7, Gemini 3.1 Pro) all returned NEEDS_FIXES. Two HIGHs reached 3-4/4 consensus: the bash-guard fail-open and the bandit camp anchor geometry bug. Three additional HIGHs were flagged by 2/4 or 1/4 reviewers but are substantive enough to warrant merge gates. Recommend a single hotfix commit on `dev` addressing H1 + H2 (plus H3 if verifiable), then re-swarm or orch-direct merge.

---

|---|---|---|---|---|
| H1 | `runtime/elders/parent/.claude/hooks/bash-guard.sh:10-16` | 4/4 | HIGH | **bash-guard fails open when `jq` is absent or hook input is malformed.** Both `jq` substitutions use `|| true`, so if `jq` is not installed or the JSON is unparseable, `tool_name` silently becomes empty, `"" != "Bash"` is true, and the hook exits 0 — allowing arbitrary Bash with no validation. This defeats the entire purpose of the guard given `permissions.defaultMode = dontAsk`. Fix: add `command -v jq >/dev/null 2>&1 || deny "jq not available"` up front; use `jq -re ... || deny "could not parse hook input"` to fail closed on bad JSON. |
| H2 | `apps/web/src/WorldMap.tsx:2300-2308` | 3/4 (Codex 5.4, Codex 5.5, Gemini, Opus 4.7 LOW) | HIGH | **`projectedBanditCampAnchor` uses region coordinates for every base instead of per-base coordinates.** `bx = region.nx * REF_W` and `by = region.ny * REF_H` are constants derived from the region, not the individual base entry. Every base in a region resolves to the same point (region center), so the "maximize distance from bases" algorithm degenerates to "maximize distance from region center." The per-base dedup set and loop are dead code. The visual fix is partial and will regress when base rendering offsets change. Fix: replace with `entry.nx * REF_W` / `entry.ny * REF_H` per loop iteration (or collapse to a single region-center check with a TODO for future per-base coordinates). |
| H3 | `packages/agents/src/cli.ts:608-621` + `bash-guard.sh:111-116` | 2/4 (Codex 5.4 HIGH, Codex 5.5 not flagged, Opus 4.7 not flagged, Gemini HIGH) | HIGH | **Bash fallback `clan submit-orders` does not validate `ELDER_N` ownership — cross-clan submission possible.** The guard only validates the file path (`/tmp/elder-N/*.json`); once validated, `runCommand()` trusts `parsed.clanId` from LLM-authored JSON and submits it unchanged. An Elder can stage `/tmp/elder-1/orders.json` with `{ "clanId": "2", ... }` and submit orders for another clan. The MCP path correctly enforces ownership via `assertOwnClanId`; the CLI fallback is the duplicate state machine that was not hardened. Fix: in the CLI `clan submit-orders` branch, derive `effectiveClanId` from `ELDER_N`, reject mismatches before forwarding to chain. |

---

## SHOULD FIX (MEDIUM)

| # | Models | Finding |
|---|---|---|
| M1 | 2/4 (Codex 5.4, Opus 4.7) | `[[ "$third" =~ ^--help\|-h$ ]]` regex binds as `(^--help)\|(-h$)` — left anchor doesn't extend past the pipe. `--help-anything` passes the help branch. Fix: `^(--help\|-h)$`. Six occurrences in bash-guard.sh. |
| M2 | 2/4 (Codex 5.4, Opus 4.7) | `settings.json` permissions hardcode `/home/claude/clan-world/...`. On non-default install paths, deny rules for `.env`/`.claude/**` no longer match real files. Fix: template `settings.json` via Makefile `sed` with actual `DEST`, or use config-relative paths. |
| M3 | 2/4 (Codex 5.4, Codex 5.5) | `post_bulletin` in `mcp.ts` lets caller supply arbitrary `slot`; CLI path always derives slot from current snapshot tick. Creates a backdating/future-dating vector through MCP. Fix: remove `slot` from MCP input schema, always use current tick. |
| M4 | 2/4 (Codex 5.4, Codex 5.5) | `projectedBanditCampAnchor` — bandit snapshot-diff state machine classifies every region change with nonzero `prev.projectedTargetClanId` as `won`, even no-defender/no-battle cases. New camp-anchor flow makes this divergence more visible. Medium-term: authoritative backend outcome flag. |
| M5 | 1/4 (Opus 4.7) | `bash-guard.sh`: `\$[A-Za-z_0-9]` regex rejects bare `$5` in bulletin bodies ("cost $5 wood"), a natural Elder communication pattern. Narrow to `\$[A-Za-z_]` (identifier-start chars only); `${`/`$(`/`$[` patterns catch the dangerous forms. |
| M6 | 1/4 (Gemini) | `packages/agents/bin/elder-mcp:17`: `exec "$TSX" ... "$ @"` — space in `"$ @"` causes POSIX shells to evaluate it as literal `$ @` instead of forwarding positional args `"$@"`. Fix: remove the space. (Verify this finding against actual file content before fixing — Gemini's path-confusion may affect reliability here.) |
| M7 | 1/4 (Codex 5.4) | `apps/clan-world-mobile/.../owner/Mwa.kt:76-82` — owner sign-in still collapses all non-fake-wallet failures to a raw error message; the new `WrongNetwork` remediation was wired through connect/mint/forge/hire/strategy/steering but not owner sign-in. |

---

## DEFER (LOW)

- **L1** (Opus 4.7): `projectedBanditCampAnchor` per-base loop is dead code — `bx/by` always equal region constants; `seen` set dedups to one point. Either inline as `.some(...)` or add a TODO. Low risk, clarity improvement.
- **L2** (Opus 4.7): `mcp.ts:319` entry-point guard `endsWith('/mcp.ts')` won't fire for compiled `dist/mcp.js`. Add `/mcp.js` to the check or use `import.meta.url`-based detection.
- **L3** (Opus 4.7): `MwaClient.kt` `WrongNetwork` string matchers are Phantom-specific English; Solflare/Backpack may use different phrasing. Acceptable for v2.4.x; file follow-up for SDK-level cluster error type when available.
- **L4** (Opus 4.7): `Makefile install` silently deletes per-elder `settings.json` overrides on every run, contradicting the "idempotent, never clobbers" README promise. Add console warning or separate `install-clean` target.
- **L5** (Codex 5.5): `MwaClusterTest` only proves `ClanWorldMwaCluster == Solana.Devnet` — doesn't prove call sites use the constant. Future direct Mainnet assignment would still pass. Add factory-based test seam.
- **L6** (Codex 5.4): `apps/web/src/WorldMap.tsx:2936-2939` — missing-target fallback uses `fromAnchor + offset` (now the camp anchor) rather than a fixed neutral region anchor; breaks spatial consistency. Minor visual regression.
- **L7** (Opus 4.7 M3): Elder skill docs don't surface `elder bulletin post` or `mcp__elder__*` tools. Elders won't discover MCP path. Add 1-para entry to `elder-base-context` skill listing MCP tools with examples.

---

- **Gemini H3 (idempotency gap in MCP submit_orders):** Gemini claims the MCP handler omits "stateful idempotency" checks for duplicate tick submissions. Codex 5.4 flagged only partial payload shape validation (M-tier). The phase goal "idempotent submit_orders validation" refers to the Convex mutation being idempotent server-side, not a client-side dedup gate. The PR description and test coverage suggest this is the intended design. Not a merge blocker — would need deeper chain-layer inspection to confirm. File as follow-up investigation.
- **Codex 5.4 H2 (shared `/tmp/elder-*/**` write access):** The settings.json grants `Write(/tmp/elder-*/**)` to all elders. Codex 5.4 flags cross-elder file tampering. This is a pre-prod system with known isolation gaps (per `project_dobox_secrets_manager_gap.md`); the per-elder settings.json path fix (M2 above) partially addresses it. Full per-elder scoping is a Phase 13+ hardening item, not a v2.4.1 blocker.

---

## Per-model verdicts

- **Codex 5.3:** ✗ prompt-echo — no review produced (hit `packages/agents/src/mcp.ts` not found sandbox error before writing output)
- **Codex 5.4:** ✓ NEEDS_FIXES — 2 HIGH (bash-guard fail-open, CLI ownership bypass), 4 MEDIUM, 3 LOW
- **Codex 5.5:** ✓ NEEDS_FIXES — 1 HIGH (bash-guard fail-open), 2 MEDIUM (bulletin slot, bandit geometry), 0 LOW
- **Opus 4.6:** ✗ 0 bytes — 3 attempts, model consistently unavailable (Claude MAX OAuth issue)
- **Opus 4.7:** ✓ NEEDS_FIXES — 1 HIGH (bash-guard fail-open), 5 MEDIUM/LOW (regex anchor, $N regex, docs gap, mcp.ts entry guard, WrongNetwork strings, Makefile idempotency)
- **Gemini 3.1 Pro:** ✓ NEEDS_FIXES — 3 HIGH (bash-guard command injection, bandit geometry, MCP idempotency), 1 MEDIUM (elder-mcp `$ @` typo), solid cross-cutting on Android normalization

---

## Cross-model overlap stats

- **H1 (bash-guard fail-open):** 4/4 — STRONG consensus, merge blocker
- **H2 (bandit camp geometry uses region not per-base coords):** 3/4 HIGH + 1/4 LOW — STRONG consensus
- **H3 (CLI submit-orders ownership bypass):** 2/4 — MEDIUM signal, recommend fix
- **M1 (regex anchor `--help|-h`):** 2/4 — confirmed independent by Codex 5.4 + Opus 4.7
- **M3 (MCP bulletin slot backdating):** 2/4 — confirmed by Codex 5.4 + 5.5
- **Gemini `$ @` typo:** 1/4 — unverified (Gemini had path confusion issues); verify before acting

---

## Decision

**Recommend Path A: single R1 fix-round on `dev` before merging.**

Scope:
1. `bash-guard.sh`: fail-closed on jq absence + malformed input (H1) — ~20 min
2. `WorldMap.tsx projectedBanditCampAnchor`: use `entry.nx/ny` not `region.nx/ny` (H2) — ~10 min
3. `cli.ts clan submit-orders`: derive `clanId` from `ELDER_N`, reject mismatch (H3) — ~20 min
4. `bash-guard.sh` regex anchor fix `^(--help|-h)$` (M1) — ~5 min
5. Add `bashGuard.test.ts` cases: jq-absent, malformed JSON, `--help-anything`, `"cost $5 wood"` (Opus 4.7 M2 cross-validates M5) — ~15 min

Total fix-round budget: ~70 min codex.

Defer to follow-up: M2 (settings.json path templating), M3 (MCP bulletin slot), M6 (elder-mcp `$ @` — verify first), M7 (owner sign-in WrongNetwork), all L-tier items.

**Liam decision needed:** Is H3 (CLI ownership bypass) a v2.4.1 merge blocker or defer-to-phase-13? The CLI path is documented in the Elder README as a fallback; if no Elder is currently using the CLI path in prod, this could be deferred. If Elders are actively using `clan submit-orders` via Bash, it is a blocker.

Issue #240, opened 2026-05-12
Android: gold faucet mint fails with 'Transaction failed to sanitize accounts offsets correctly'


claude-do:
Reported 2026-05-11 during v2.4.1 demo bring-up. Android app shows error "Something went wrong — invalid transaction: Transaction failed to sanitize accounts offsets correctly" when user taps "mint 100k gold" via faucet.

**Root cause (Claude diagnosis, confidence HIGH):**

The Android client builds an ATA-create-idempotent instruction in \`GoldSolanaClient.kt:184-195\` where the user's pubkey appears in two slots with conflicting flags:
- slot 0: \`payer\` — \`isSigner=true, isWritable=true\`
- slot 2: \`owner\` — \`isSigner=false, isWritable=false\`

For self-funded ATAs, \`payer == owner\` (line 108: \`createAssociatedTokenAccountIdempotent(owner, owner, mint, ownerAta)\`).

The \`com.solanamobile:web3-solana-jvm:0.3.0-beta4\` \`Message.Builder.build()\` does NOT union privilege flags across slots — it bucket-adds the pubkey to BOTH \`writableSigners\` AND \`readOnlyNonSigners\` independently. The resulting compiled message has the owner pubkey twice in the account-key array (signer partition + readonly partition). Solana RPC's pre-flight sanitizer detects the privilege conflict and rejects with the exact error.

**Secondary bug (also in GoldSolanaClient.kt):**

Line 149: \`Message.Builder(instructions.toMutableList(), payer)\` uses positional args. The data-class signature is \`Builder(instructions, blockhash, feePayer)\` — so \`payer\` lands in the \`blockhash\` slot. \`setRecentBlockhash\` overwrites blockhash later, but \`feePayer\` is left null. Should use fluent API: \`Message.Builder().addFeePayer(payer).setRecentBlockhash(blockhash).addInstruction(it)...\`

**Recommended fixes:**

1. **Primary (load-bearing for sanitize error):** When \`payer == owner\` in ATA-create, promote owner slot's flags so both occurrences land in the same bucket:
   \`\`\`kotlin
   val ownerIsSelf = owner.bytes.contentEquals(payer.bytes)
   // ...
   AccountMeta(owner, isSigner = ownerIsSelf, isWritable = ownerIsSelf),
   \`\`\`
[...]

PR #241, opened 2026-05-12
fix(android): faucet ATA self-owned dedup + Builder fluent API

claude-do:
Closes #240

## Summary

Two bugs in `GoldSolanaClient.kt` combined to produce "Transaction failed to sanitize accounts offsets correctly" when minting 100k GOLD from the Android faucet:

1. **ATA-create AccountMeta flag mismatch (primary).** For self-funded ATAs (`payer == owner`, e.g. faucet claim) the same pubkey was emitted in slot 0 as `(isSigner=true, isWritable=true)` and slot 2 as `(isSigner=false, isWritable=false)`. `web3-solana:0.3.0-beta4`'s `Message.Builder.build()` bucket-adds each `AccountMeta` to one of four `LinkedHashSet`s (writable-signers / readonly-signers / writable-non-signers / readonly-non-signers) without flag-unioning, so the owner pubkey landed in TWO buckets and appeared twice in `message.accounts`. Solana's pre-flight sanitizer rejects duplicates. Fix: when `payer == owner`, promote the owner-slot flags so both slots route to the same bucket and collapse to one entry.

2. **`Message.Builder` constructor positional misalignment (secondary).** `Message.Builder(instructions.toMutableList(), payer)` placed `payer` into the `blockhash` slot — `setRecentBlockhash` immediately overwrote it, so functionally harmless in this SDK (Builder has only `instructions` + `blockhash` fields; fee payer is implicit from the first writable signer), but confusing intent. Fix: use the no-arg constructor + fluent `setRecentBlockhash` + `addInstruction`.

Note: the brief sketched `addFeePayer(payer)`; the actual SDK has no such method (verified by inspecting `Message$Builder` bytecode in `web3-solana-jvm:0.3.0-beta4`). The fee payer is inferred from the first writable signer in the instruction list, which for both flows here is `owner`. The unused `payer` parameter on `buildTransaction` was therefore dropped.

## Test plan

- [x] `:app:assembleDebug` — BUILD SUCCESSFUL
- [x] `:app:testDebugUnitTest` — all tests pass, including 2 new regression tests in `GoldSolanaClientTest`:
  - `selfFundedAtaCreatePutsOwnerInAccountsExactlyOnce` — verified to FAIL when fix is reverted, PASS with fix applied
  - `separatePayerKeepsOwnerReadonly` — guards the non-self-owned (treasury skip-tax) path stays readonly
- [ ] Manual: mint 100k GOLD from Android app on devnet succeeds without sanitize error
- [ ] Manual: existing skip-tax burn flow (`treasuryOwner != owner`) still works (no regression)

### DIFF
--- a/apps/clan-world-mobile/app/src/main/java/world/clan/app/data/gold/GoldSolanaClient.kt
+++ b/apps/clan-world-mobile/app/src/main/java/world/clan/app/data/gold/GoldSolanaClient.kt
@@ -99,26 +99,7 @@ class GoldSolanaClient(
   }
 
   suspend fun buildFaucetClaim(ownerBase58: String, memo: String): ByteArray {
-    val owner = ownerBase58.toPubkey()
-    val ownerAta = associatedTokenAddress(owner, mint)
-    val mintAuthority = pda("mint-authority")
-    return buildTransaction(
-      listOf(
-        createAssociatedTokenAccountIdempotent(owner, owner, mint, ownerAta),
-        TransactionInstruction(
-          faucetProgram,
-          listOf(
-            AccountMeta(owner, isSigner = true, isWritable = true),
-            AccountMeta(mint, isSigner = false, isWritable = true),
-            AccountMeta(ownerAta, isSigner = false, isWritable = true),
-            AccountMeta(mintAuthority, isSigner = false, isWritable = false),
-            AccountMeta(TOKEN_PROGRAM_ID, isSigner = false, isWritable = false),
-          ),
-          anchorDiscriminator("global:claim"),
-        ),
-        memo(owner, memo),
-      ),
-    )
+    return buildTransaction(faucetClaimInstructions(ownerBase58.toPubkey(), memo))
   }
 
   suspend fun buildBurn(
@@ -127,7 +108,38 @@ class GoldSolanaClient(
     memo: String,
     skipTax: Long = 0L,
   ): ByteArray {
-    val owner = ownerBase58.toPubkey()
+    return buildTransaction(burnInstructions(ownerBase58.toPubkey(), amount, memo, skipTax))
+  }
+
+  internal suspend fun faucetClaimInstructions(
+    owner: SolanaPublicKey,
+    memo: String,
+  ): List<TransactionInstruction> {
+    val ownerAta = associatedTokenAddress(owner, mint)
+    val mintAuthority = pda("mint-authority")
+    return listOf(
+      createAssociatedTokenAccountIdempotent(owner, owner, mint, ownerAta),
+      TransactionInstruction(
+        faucetProgram,
+        listOf(
+          AccountMeta(owner, isSigner = true, isWritable = true),
+          AccountMeta(mint, isSigner = false, isWritable = true),
+          AccountMeta(ownerAta, isSigner = false, isWritable = true),
+          AccountMeta(mintAuthority, isSigner = false, isWritable = false),
+          AccountMeta(TOKEN_PROGRAM_ID, isSigner = false, isWritable = false),
+        ),
+        anchorDiscriminator("global:claim"),
+      ),
+      memo(owner, memo),
+    )
+  }
+
+  internal suspend fun burnInstructions(
+    owner: SolanaPublicKey,
+    amount: Long,
+    memo: String,
+    skipTax: Long = 0L,
+  ): List<TransactionInstruction> {
     val ownerAta = associatedTokenAddress(owner, mint)
     val instructions = mutableListOf<TransactionInstruction>()
     instructions += burnChecked(ownerAta, mint, owner, amount)
@@ -137,15 +149,16 @@ class GoldSolanaClient(
       instructions += transferChecked(ownerAta, mint, treasuryAta, owner, skipTax)
     }
     instructions += memo(owner, memo)
-    return buildTransaction(instructions)
+    return instructions
   }
 
   private suspend fun buildTransaction(
     instructions: List<TransactionInstruction>,
   ): ByteArray {
     val blockhash = latestBlockhash()
+    val canonical = unionAccountFlags(instructions)
     val builder = Message.Builder().setRecentBlockhash(blockhash)
-    instructions.forEach { builder.addInstruction(it) }
+    canonical.forEach { builder.addInstruction(it) }
     return Transaction(builder.build()).serialize()
   }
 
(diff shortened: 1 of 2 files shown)


claude-do:
**Tier 1 (Claude subagent)** · round 1 · head `7d9cecc`

Verdict: **NEEDS_FIXES** — 1 CRITICAL, 2 MEDIUM

## CRITICAL · multi-bucket bug remains in the full `buildFaucetClaim` flow (and `buildBurn` with skip-tax)

The fix in `createAssociatedTokenAccountIdempotent` resolves the in-instruction self-reference (slot 0 vs slot 2 of the ATA-create ix). It does **not** address the cross-instruction variant: `memo()` always emits `AccountMeta(owner, isSigner=true, isWritable=false)`, which routes owner into `set2` (readonly-signers) while ATA-create slot 0 + `claim` slot 0 route owner into `set1` (writable-signers). Result: compiled `LegacyMessage` has `signatureCount=1, readOnlyAccounts=1` → **0 writable signers**, owner labeled readonly-signer.

I verified this empirically: with the PR's fix applied, a test that constructs the full `[ataCreate, claim, memo]` instruction list and inspects the compiled message via reflection on `getSignatureCount`/`getReadOnlyAccounts` reports `writableSigners=0`. Same pattern in `buildBurn` when `skipTax > 0` (`burnChecked` + `createAssociatedTokenAccountIdempotent(owner, treasuryOwner, ...)` + `transferChecked` + `memo` → owner in both set1 and set2).

The on-chain symptom changes from "sanitize accounts offsets correctly" to a writable-account-mismatch / fee-payer-not-writable error, but the faucet claim will still fail. The two manual smoke-test items in the PR description are precisely what would have caught this — both still unchecked.

I also confirmed via `javap -c` on `web3-solana-jvm:0.3.0-beta4` that:
- `Message.Builder.build()` uses 4 separate `LinkedHashSet` buckets with no flag-union.
- `accounts` IS deduped (via `SetsKt.plus` → `LinkedHashSet.addAll` → equals-by-byte-array). So the PR description's "appeared twice in `message.accounts`" framing is slightly off — the duplicate is in the *header count* fields, not in `accounts`.
- `signatureCount = (set1 ∪ set2).size` (post-dedup); `readOnlyAccounts = set2.size` (pre-dedup) ← this asymmetry is the actual root cause.

**Recommended fix:** pre-pass `buildTransaction` to union per-pubkey flags across all `AccountMeta`s before adding to `Message.Builder`. This subsumes the current helper-level patch and handles every cross-instruction variant. Full kotlin sketch in the findings doc.

## MEDIUM · the new unit test's 3 of 4 assertions are vacuous

When I reverted just the helper-level patch and ran `:app:testDebugUnitTest`, the test failed at the **4th** assertion (`ownerSlot.isSigner && ownerSlot.isWritable`) — a tautology asserting what the fix literally wrote. The other 3 (`message.accounts.count == 1`, `distinctKeys.size == accounts.size`, `payerSlot is writable signer`) all pass with the fix reverted, because the SDK already dedups `accounts`. The meaningful invariant to assert is the compiled-message header consistency, not the `accounts.count` of a single-instruction transaction.

## MEDIUM · manual devnet smoke (unchecked in PR) is the actual sign-off gate

Given the residual CRITICAL above, please do not merge until end-to-end faucet claim on devnet returns a confirmed signature. Or fix the union-flag root cause and re-smoke.

## Non-issues verified

- Anchor `Claim` constraints unaffected by ATA-create flag promotion (different program).
- `Message.Builder()` no-arg + fluent change is correct — bytecode confirms Builder has only `instructions` + `blockhash` fields, no `feePayer`.
- `payer` parameter removal from `buildTransaction` — only 2 callers, both updated.

Full findings: `/tmp/swarm-review-tier1-241.md` (local).

#claude-subagent

claude-do:
## Round 2 fix — multi-bucket bug closed at the right layer

**Head:** 799320b
**Reviewer findings addressed:** Tier 1 CRITICAL + Tier 2 [SEVERITY: high] (both caught the same thing — round-1 fix only covered the ATA instruction in isolation; the full `buildFaucetClaim` path still split owner across writable-signer and readonly-signer buckets because `memo()` emits owner as readonly-signer).

### What changed in production code

`GoldSolanaClient.unionAccountFlags(instructions)` — new internal pre-pass that runs before every `Message.Builder.build()`. It computes the OR-union of `(isSigner, isWritable)` per pubkey across ALL instructions and normalizes every occurrence to that union. Routes through the single `buildTransaction` chokepoint, so it covers `buildFaucetClaim`, `buildBurn(skipTax > 0)`, and any future transaction shape automatically.

This matches what canonical Solana SDKs do (Rust `solana-sdk` `Message::new`, JS `@solana/web3.js` `Transaction.compileMessage` — both union flags before bucketing). `com.solanamobile:web3-solana-jvm:0.3.0-beta4` does not, so we compensate at the caller layer.

The round-1 narrow flag-promotion inside `createAssociatedTokenAccountIdempotent` is now redundant and was removed (per Tier 1's recommendation — symptom-fix in one helper while the pattern remained elsewhere was the asymmetry Tier 1 flagged).

`buildFaucetClaim` and `buildBurn` are now thin wrappers around `faucetClaimInstructions` / `burnInstructions` (new `internal` helpers) so tests can compile the full instruction list without RPC.

### What changed in tests

Old test file had 3 vacuous assertions Tier 1 identified (the dedup-tautology + flag-tautology + writable-flag-tautology — all passed even when the fix was reverted). Dropped them.

New tests assert the **canonical Solana message invariant** rather than tautologies:

```
accounts.size == signatureCount + writableNonSigners + readOnlyNonSigners
```

i.e. `writableNonSigners = accounts.size - signatureCount - readOnlyNonSigners` must be `>= 0`. Before the union pre-pass, a pubkey landing in 2 buckets could push the raw bucket-sum past the deduped `accounts.size`, making `writableNonSigners` negative — that's the exact mismatch the on-chain sanitizer trips on.

7 tests total:

| Test | Asserts |
|---|---|
| `selfFundedAtaCreateOwnerEndsUpWritableSignerInCompiledMessage` | Single-ix happy path: owner lands in writable-signer slice |
| `separatePayerKeepsOwnerReadonlyNonSigner` | Owner ≠ payer: owner correctly in readonly-non-signer suffix |
| `faucetClaimFullInstructionListHasConsistentHeader` | **The critical one.** Full ATA-create + claim + memo: header partition is self-consistent + no duplicate pubkeys + owner in writable-signer bucket |
| `faucetClaimUnionsOwnerFlagsAcrossMemoAndPayerSlots` | Sanity-checks the input shape (memo emits readonly-signer, ATA emits writable-signer) AND that the union normalizes every owner occurrence to writable-signer |
| `burnWithSkipTaxHasConsistentHeader` | Full `buildBurn(skipTax=5)`: header consistency + owner in writable bucket |
| `burnWithoutSkipTaxStillProducesConsistentHeader` | Skip-tax-0 burn: owner stays readonly-signer (union is a no-op here) — verifies no regression on the working path |
| `unionAccountFlagsIsIdempotentOnAlreadyCanonicalInput` | Applying union twice = applying it once |

### Regression-proofed

I temporarily replaced `unionAccountFlags` with `return instructions` (no-op) and re-ran the suite — **3 of 7 tests fail** on the full-faucet, burn-with-skip-tax, and post-union owner-shape assertions. The 4 remaining tests still pass because their inputs don't trigger multi-bucket splits, which is the correct behavior. This proves the new tests are catching the real invariant, not asserting tautologies.

### Test command + result

```
cd apps/clan-world-mobile
ANDROID_HOME=/opt/android-sdk \
  CLAN_WORLD_MAP_URL=https://app.clan-world.com \
  CLAN_WORLD_TERMINAL_BASE_URL=https://cockpit.clan-world.com \
  ./gradlew :app:testDebugUnitTest
```

```
BUILD SUCCESSFUL in 4s
GoldSolanaClientTest: 7 tests, 0 failures, 0 errors
(Full suite: FakeWalletPolicyTest 5/5, MwaClusterTest 2/2, WalletNameServiceTest 4/4,
 ConvexJsonNumberTest 11/11, GoldSolanaClientTest 7/7 — 29 total, 0 fail.)
```

### Not addressed (Tier 1 MEDIUM #2 — informational)

Tier 1 MEDIUM #2 suggested making `createAssociatedTokenAccountIdempotent` a top-level / companion-object helper so tests don't need a `GoldSolanaClient` instance. Deferred — the test does pass a real instance with valid placeholder base58 (system-program pubkey, decodes to 32 bytes), so the eager-decode concern doesn't bite today. Refactor can land in a follow-up if/when more tests need similar pattern.

Tier 1 MEDIUM #3 (unchecked manual test plan checkboxes in PR description) — still applies; devnet smoke needs to confirm faucet-claim end-to-end transaction confirmation. The union pre-pass should resolve the sanitize error per the empirical reproduction in the Tier 1 findings, but a real on-device test is what gets the checkboxes ticked.

### References
- Tier 1 findings: `/tmp/swarm-review-tier1-241.md`
- Tier 2 findings: `/tmp/swarm-review-tier2-241.md` (high-severity at line 1582)

claude-do:
**Claude subagent (Tier 1)** · round 2 · head `799320b`

Verdict: CLEAN (0 HIGH, 0 MED, 2 LOW)

Empirically verified R2's regression-proof claim: temporarily no-op'd `unionAccountFlags` (replaced body with `return instructions`), re-ran tests — exactly the 3 named tests fail (`faucetClaimFullInstructionListHasConsistentHeader`, `burnWithSkipTaxHasConsistentHeader`, `faucetClaimUnionsOwnerFlagsAcrossMemoAndPayerSlots`); other 4 pass as expected (single-ix or no-multi-bucket). Restored impl → all 7 green. Working tree clean.

Validated: per-pubkey OR-union; `programId` / `data` preservation; per-instruction account order preservation; sole `buildTransaction` sink routes both faucet-claim + burn paths through pre-pass. No other multi-bucket vectors in this client surface.

LOW (style, non-blocking):
- `GoldSolanaClient.kt:243` — `union[…]!!` non-null assert; `getValue` or `?: error(...)` clearer.
- `GoldSolanaClient.kt:228` — `internal fun unionAccountFlags` exposed for tests; consider `@VisibleForTesting`.

#claude-subagent

PR #242, opened 2026-05-12
fix(web): bandit animation polish — base-anchored circle, +50% diameter, solid 3rd tombstone, red glow

claude-do:
## Summary

Four bandit-animation UX bugs reported by Liam after the 2026-05-11 demo session, all confined to `apps/web/src/WorldMap.tsx`:

### 1. Circle anchor was wrong
When the bandit attack reached the **circling** phase, the orbit center was set to a `fromAnchor + (40, 30)` offset from the bandit's camp — an arbitrary empty patch of map. This happened because the `defeated` branch of `BanditDiffOutcome` carried no `targetClanId`, so `renderBattlePhase` couldn't find the target base and fell through to the empty-offset fallback.

**Fix:** add `targetClanId: string | null` to the `defeated` variant. Propagate it from:
- the snapshot-diff path (`useEffect` on snapshot) — read `prev.projectedTargetClanId`.
- the live-bandit `BanditState.Defeated` synth path inside `computeBanditAnimPhase`.

`renderBattlePhase` now reads `phase.outcome.targetClanId` for **both** `won` and `defeated`, looks up the rendered base by clan id, and anchors `targetCenter` to `targetBase.container.x / .y - 30*scale`. Two safer fallbacks replace the old empty-offset one:
- any rendered base in the bandit's region.
- the region's projected geographic anchor (`projectedRegionAnchor`).

### 2. Circle diameter +50%
Constant at line 2958 (now ~3062) bumped: `38 * layoutScale * (1 - circleT * 0.4)` → `57 * layoutScale * ...`. The `* 0.55` y-squish (ellipse for perspective) is unchanged.

### 3. Tombstone (3rd death frame) stops flashing
The death animation cycles through 3 sprites: 0 = back-knockdown, 1 = face-down, 2 = tombstone (held during the 8-second fade-out). The 8 Hz square-wave flicker was being applied to **all** frames including the tombstone, so the tombstone strobed during its fade.

**Fix:** once `deathFrameIdx >= 2`, `flickerAlpha` is locked to `1`. Alpha is then driven only by `(1 - fadeT)`.

### 4. Red glow on all 3 bandit sprite states
Standing / moving / attacking all share the same `walkers: Sprite[3]`, so a single halo array covers them. Death sprites get their own halo array.

**Approach** (since `pixi-filters`'s `GlowFilter` is **not** in the `apps/web` dep tree, only `pixi.js@8.18.1`):
- For each walker / death sprite, pre-create a paired "glow" sprite added to `worldDynamic` **before** the main sprite (z-order behind).
- Glow sprite: `tint = 0xff2222`, `alpha = 0.7`, `BlurFilter({ strength: 6, quality: 2 })`.
- A single `syncBanditGlows()` helper mirrors texture / position / scale / visibility / zIndex (`-1` so it always sits below).
- `updateBanditAnimation` wraps its inner body in `try { } finally { syncBanditGlows() }` so every early-return phase still gets a synced halo.

Changes also extend the `banditAnimRef` shape (+`walkerGlows`, `deathGlows`) and the unmount-reset blob.

## Files changed

- `apps/web/src/WorldMap.tsx` (+127 / −14)

## Verification

- `pnpm --filter @clan-world/web typecheck` — **no new errors in WorldMap.tsx**. The repo's pre-existing tsc errors in `apps/server/convex/*`, `EventTicker.tsx`, and `apps/web/src/components/cockpit/*` are present on `dev` head as well (confirmed by stashing + retrying).
- Lint stub is a no-op (`echo 'lint stub'`).
- `pnpm --filter @clan-world/web build` also fails on the same pre-existing convex/ts errors, irrespective of this PR.

## Test plan

- [ ] **Visual UAT (Liam):** spawn a bandit, let it attack a base. Confirm:
  - [ ] The 3 bandits circle the **rendered base sprite**, not a patch of empty map.
  - [ ] Circle diameter is noticeably wider than before (~50% larger).
  - [ ] During the 8s tombstone fade, the 3rd sprite (gravestone) does **not** flash — it stays solid and fades cleanly to 0.
  - [ ] All 3 bandit sprite states (camp/standing in `camp_telegraph`, marching in `battle.march`, circling in `battle.circle`, traversing in `no_battle_advance`) show a soft red halo around them.
- [ ] Smoke: `apps/web/tests/e2e/02-demo-mode.spec.ts` still passes (DEMO_MODE path unchanged — these fixes only affect the live `phase` machine).

**Needs visual UAT** — no e2e covers the animation visuals.

### DIFF
--- a/apps/web/src/WorldMap.tsx
+++ b/apps/web/src/WorldMap.tsx
@@ -1,6 +1,6 @@
 import { useEffect, useMemo, useRef, useState } from 'react';
 import { useSafeQuery as useQuery } from './hooks/useSafeQuery';
-import { Application, Assets, ColorMatrixFilter, Container, Graphics, Rectangle, Sprite, Text } from 'pixi.js';
+import { Application, Assets, BlurFilter, ColorMatrixFilter, Container, Graphics, Rectangle, Sprite, Text } from 'pixi.js';
 import { Viewport } from 'pixi-viewport';
 import { useAgentLogs, type AgentLog } from './useAgentLogs';
 import { WorldNoticePanel } from './WorldNoticePanel';
@@ -574,7 +574,7 @@ type CombatTrigger = {
 // "lastOutcome" memo. T3 last 10s = telegraph. T4 = full battle / advance / escape.
 
 type BanditDiffOutcome =
-  | { type: 'defeated'; resolvedTick: number; fromRegion: number }
+  | { type: 'defeated'; resolvedTick: number; fromRegion: number; targetClanId: string | null }
   | { type: 'won'; resolvedTick: number; fromRegion: number; toRegion: number; targetClanId: string | null }
   | { type: 'no_battle_advance'; resolvedTick: number; fromRegion: number; toRegion: number };
 
@@ -902,10 +902,17 @@ export function computeBanditAnimPhase(
 
       const fromRegionKey = regionKeyByChainId[lastOutcome.fromRegion];
       if (!fromRegionKey) return { kind: 'hidden' };
+      // Both 'won' and 'defeated' outcomes now carry a targetClanId so the
+      // attack-circle animation anchors to the actual rendered base sprite
+      // instead of a stale fromAnchor offset (empty patch of map).
+      const phaseTargetClanId =
+        lastOutcome.type === 'won' || lastOutcome.type === 'defeated'
+          ? lastOutcome.targetClanId
+          : null;
       return {
         kind: 'battle',
         regionKey: fromRegionKey,
-        targetClanId: lastOutcome.type === 'won' ? lastOutcome.targetClanId : null,
+        targetClanId: phaseTargetClanId,
         outcome: lastOutcome,
         battleT,
       };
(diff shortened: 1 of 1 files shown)


claude-do:
**Claude subagent (Tier 1)** · round 1 · head `bbcb2fa`

Verdict: **CLEAN**

Validated:
- `targetClanId` correctly flows through BOTH `defeated` paths (snapshot-diff line 1267-1273 + live synth line 949-963). Discriminated-union narrowing at the render site checks both `'won' || 'defeated'` before reading `outcome.targetClanId`. Exhaustiveness preserved.
- `(- 30*scale)` y-offset: base sprites are anchored at `(0.5, 1)` (bottom-center), so the offset correctly places the orbit above the base. Matches prior usage; only the fallback chain changed (and improved — old `fromAnchor + (40,30)*scale` empty-patch fallback replaced with regionBase → projectedRegionAnchor → fromAnchor cascade).
- Glow lifecycle: per-sprite `BlurFilter` instances (no shared filter contention), added as `worldDynamic` children, cleaned up via `app.destroy(true)` recursive teardown. No Pixi leak. Reset arrays at mount + unmount.
- `syncBanditGlows`: fixed-size arrays (3+3), never mutated mid-render, bounds-safe. Hidden-source early-return avoids stale z/position drift.
- BlurFilter API matches Pixi v8.18.1 (`{ strength: 6, quality: 2 }` verified against `BlurFilterOptions`).
- Z-order: `worldDynamic.sortableChildren = true` confirmed (line 783); glow at `walker.zIndex - 1` works correctly. addChild-order comment is slightly stale but cosmetic.
- Tombstone flicker gating (`deathFrameIdx >= 2 ? 1 : flickerOn`) is off-by-one safe — death-texture access is clamped via `Math.min(textures.length - 1, ...)` in `applyDeathFrame`.
- `tsc --noEmit` shows zero errors in `WorldMap.tsx`. Pre-existing `apps/server/convex/**` errors are unrelated.

Non-blocking nits (don't gate merge):
- Comment at line 2748-2749 ("Glow added FIRST so it sits behind") is misleading given `sortableChildren = true` — zIndex is what matters, not insertion order. Consider tightening.
- No unit tests for `computeBanditAnimPhase`; symmetric with prior state.

#claude-subagent

Issue #251, opened 2026-05-12
v2.5.2: BlurFilter instances not destroyed on Pixi teardown — HMR leak


claude-do:
Super-swarm finding from Opus 4.7 on PR #246. Logged for v2.5.2.

apps/web/src/WorldMap.tsx:2891-2899 — each of 3 walker glows + 3 death glows holds its own BlurFilter. Cleanup path resets the arrays but doesn't .destroy() the filters. PixiJS Sprite.destroy() doesn't cascade into sprite.filters[]. Each BlurFilter allocates an FBO on first use — on vite HMR cycles these leak.

**Fix**: in cleanup block (useEffect return / app.destroy path), iterate the glow arrays and filter.destroy() each, OR call sprite.destroy({ children: true, filters: true }) (Pixi v8 supports the option).

Production users won't hit hot-reload, leak is bounded to ~6 filters per page-load, but dev experience is worse.

Tier: MEDIUM (dev-only)
Source: Opus 4.7 super-swarm review on PR #246

Issue #250, opened 2026-05-12
v2.5.2: prefers-reduced-motion sampled only at init time


claude-do:
Super-swarm finding from Opus 4.7 + Gemini 3.1 Pro on PR #246. Logged for v2.5.2.

`apps/web/src/WorldMap.tsx:1859-1867` — `matchMedia('(prefers-reduced-motion: reduce)').matches` is sampled once at app-init. If user toggles OS motion preference mid-session, winter snow continues animating (or never starts).

Matches local convention in this codebase, but ideally should subscribe to MediaQueryList's `change` event.

**Fix**: register a change listener that routes through `winterSnowRef.current?.setActive(...)`, creating the handle lazily on first activation if init was reduced.

Tier: LOW
Source: Opus 4.7 + Gemini 3.1 Pro super-swarm review on PR #246

Issue #249, opened 2026-05-12
v2.5.2: dead-clansman halo lazy-create on revive


claude-do:
Super-swarm finding from codex 5.5 + codex 5.4 + Opus 4.7 on PR #246. Logged for v2.5.2.

`apps/web/src/WorldMap.tsx:2284` — `makeLiveClansmanMarker` only allocates `halo: Graphics` if `missionActive && !isDead`. So if a clansman was first seen DEAD and is later revived WITH an active mission, `existing.halo` is null and the dead→alive transition's `if (existing.halo) existing.halo.visible = marker.missionActive` is a no-op — no halo will ever appear.

PR #244 R2 added the symmetric `applyAliveVisualState` helper but didn't address the lazy-creation gap. The R2 comment promises behavior the code can't deliver.

**Fix**: in the dead→alive branch, lazily build the halo when needed:
```ts
if (!existing.halo && marker.missionActive) {
  existing.halo = makeHaloGraphics();
  existing.node.addChildAt(existing.halo, 0);
} else if (existing.halo) {
  existing.halo.visible = marker.missionActive;
}
```
Apply the same lazy-create pattern to idle→active transitions to close the pre-existing gap.

Tier: MEDIUM
Source: codex 5.5 + codex 5.4 + Opus 4.7 super-swarm review on PR #246

Issue #248, opened 2026-05-12
v2.5.2: battle-attack fallback picks wrong clan base when target missing


claude-do:
Super-swarm finding from codex 5.4 on PR #246 (v2.5.0 release). Logged for v2.5.2.

`apps/web/src/WorldMap.tsx:3186` — when `targetClanId` is no longer present in snapshot, the bandit-circling animation falls back to `drawn.bases.find(b => b.clan.homeRegion === phase.regionKey)`. In a multi-clan region, this picks a SURVIVING clan that wasn't attacked. UI shows bandits circling the wrong clan = false game state.

This is worse than the old 'empty patch' fallback because it communicates incorrect state, not just imprecise placement.

**Fix**: if the specific target base is missing, fall back to a neutral region anchor only; do NOT choose an arbitrary base from the same region.

Tier: MEDIUM
Source: codex 5.4 super-swarm review on PR #246

PR #253, opened 2026-05-12
feat(v2.5.2): super-swarm deferred fixes (#248 #249 #250 #251)

claude-do:
## Summary

Implements the four super-swarm v2.5.0 findings that were deferred out of the v2.5.1 hotfix release. All four are in `apps/web/src/WorldMap.tsx`. No new dependencies, no schema changes, no production-data impact — purely client-side rendering / lifecycle fixes.

Closes #248, closes #249, closes #250, closes #251.

## Fix 1 — #248: Battle fallback picks wrong clan's base (codex 5.4 MED)

**File:** `apps/web/src/WorldMap.tsx` (battle-phase fallback block, ~line 3186).

When `targetClanId` from the `BanditDiffOutcome.defeated` variant was missing from the snapshot, the v2.5.0 fallback was:

```ts
const regionBase = drawn.bases.find(b => b.clan.homeRegion === phase.regionKey);
```

In a multi-clan region this would pick an **arbitrary same-region base** — visually attributing the bandit attack to a clan that was never actually targeted. Bandits would circle a *surviving* clan's base = false game state.

**Fix:** Remove the "any same-region base" tier entirely. Skip directly to the neutral `projectedRegionAnchor(phase.regionKey)` — the region's geographic center — which reads as "battle is happening somewhere in this region" without indicting a specific surviving clan.

## Fix 2 — #249: Dead-clansman halo lazy-create on revive (codex 5.5 + 5.4 + Opus 4.7 MED)

**File:** `apps/web/src/WorldMap.tsx` (~lines 2470-2545 in sync loop, plus a small helper extraction ~line 2337).

`makeLiveClansmanMarker` only allocates `halo: Graphics` if `missionActive && !isDead`. So a clansman first observed DEAD never had a halo allocated. The v2.5.0 R2 revive path was:

```ts
if (existing.halo) existing.halo.visible = marker.missionActive;
```

When `existing.halo` was `null` (first-seen-dead case), this was a silent no-op — the revived clansman never grew a halo even though `missionActive=true`.

**Fix:**
- Extract the halo geometry into `makeHaloGraphics()` so both initial-create and sync-loop lazy-create use identical settings (ring radius 16, color `0xffe6a0`, stroke 2px alpha 0.95).
- On dead→alive transition: if `!existing.halo && marker.missionActive`, allocate via `makeHaloGraphics()` + `addChildAt(_, 0)` (z-order behind body, matches `makeLiveClansmanMarker`).
- **Same pattern applied to idle→active transition for alive markers** — captured via `wasMissionActive = existing.missionActive` BEFORE `Object.assign` overwrites it. Without this, an alive+idle clansman who gets a new mission would also lack a halo until full re-mount.

## Fix 3 — #250: prefers-reduced-motion sampled only at init (Opus 4.7 + Gemini LOW)

**File:** `apps/web/src/WorldMap.tsx` (~lines 1859-1910 init + cleanup ~line 1947).

`window.matchMedia('(prefers-reduced-motion: reduce)').matches` was sampled exactly once at app-init. Toggling the OS motion preference mid-session had no effect — snow kept animating (or never appeared, in the inverse case).

**Fix:** Subscribe to `MediaQueryList`'s `change` event. The handler:
- On OFF→ON (user enables reduced-motion): `winterSnowRef.current?.setActive(false)` if the system was built (keeps handle for re-enable; no re-allocation cost).
- On ON→OFF (user disables reduced-motion): **lazily build the snow handle** if init skipped it — `createWinterSnow` + `app.stage.addChild` + restore snapshot-driven active state via `(snapshotRef.current as any)?.winterActive === true`.

Listener + MediaQueryList stored on `reducedMotionMqRef` / `reducedMotionHandlerRef`. Cleanup in the pixi-init `useEffect` return path removes the listener.

## Fix 4 — #251: BlurFilter not destroyed on Pixi teardown (Opus 4.7 MED, dev-only)

**File:** `apps/web/src/WorldMap.tsx` (cleanup block ~line 1975, before `banditAnimRef.current = {...}` reset).

Each of 3 walker glows + 3 death glow sprites holds its own `BlurFilter` (instantiated in `makeGlow` at ~line 2901). Each `BlurFilter` allocates a render-target framebuffer. The v2.5.0 cleanup reset the arrays (`walkerGlows: []`) but did NOT destroy the filters.

Verified against `@types/pixi.js@8.18` (`destroyTypes.d.ts`): `DestroyOptions` only exposes `children` / `texture` / `textureSource` / `context` / `style` — **no `filters` flag**. `Sprite.destroy()` does NOT cascade into `sprite.filters[]`. Result: FBO leak across every Vite HMR cycle in dev, plus StrictMode double-mount in prod builds.

**Fix:** Iterate `[...walkerGlows, ...deathGlows]` BEFORE the array reset, explicitly call `filter.destroy()` (inherited from `Shader` per pixi v8's class hierarchy), then `glow.filters = null`. Wrapped in try/catch with `import.meta.env.DEV` debug log — the StrictMode double-cleanup pattern can otherwise spam errors.

## Verification

- `pnpm --filter @clan-world/web build` — **GREEN** (7.29s, no new warnings)
- `tsc -b` clean — no new TS errors anywhere
- Diff stat: **1 file changed, 141 insertions(+), 22 deletions(-)** (WorldMap.tsx only)

## Test plan

- [ ] Smoke: load `/world` in dev, no console errors
- [ ] #248: trigger a bandit battle in a multi-clan region with a missing targetClanId in the snapshot (force via devtools state hack), confirm orbit ring centers on region geo anchor, NOT a surviving clan's base
- [ ] #249: kill a clansman, then revive on a tick where their clan has an active mission — halo ring should appear behind their body
- [ ] #249b: assign a fresh mission to an alive+idle clansman — halo should appear without remount
- [ ] #250: toggle OS reduced-motion mid-session both directions — snow should freeze/resume accordingly
- [ ] #251: dev-only HMR sanity — repeatedly save WorldMap.tsx in dev mode; GPU memory in Chrome DevTools should stay flat (no FBO accumulation)

### DIFF
--- a/apps/web/src/WorldMap.tsx
+++ b/apps/web/src/WorldMap.tsx
@@ -1153,6 +1153,12 @@ export function WorldMap() {
   // React DOM UI). Created post-init; activated/deactivated via setActive in a
   // separate effect that watches `snapshot.winterActive`.
   const winterSnowRef = useRef<WinterSnowHandle | null>(null);
+  // prefers-reduced-motion media query + change listener. Sampled once at app
+  // init and re-checked on any OS-level toggle (issue #250). Stored on refs so
+  // the cleanup in the pixi-init useEffect can remove the listener without
+  // re-running on every render.
+  const reducedMotionMqRef = useRef<MediaQueryList | null>(null);
+  const reducedMotionHandlerRef = useRef<((event: MediaQueryListEvent) => void) | null>(null);
 
   // Bandit attack animation (docs/planning/bandit-animation-impl-plan.md).
   // Snapshot-diff tracking: derive last resolution outcome on every snapshot tick.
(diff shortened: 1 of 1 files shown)


claude-do:
**Claude subagent (Tier 1)** · round 1 · head `8ad49c3`

Verdict: **CLEAN** (1 LOW non-blocking, pre-existing & out of scope)

All four super-swarm fixes validated:

- **#248** — `projectedRegionAnchor` (L2567) is neutral region geometry, zero clan coupling. Final `?? fromAnchor` fallback at L3318 catches the null-region case. No false-attribution surface remains.
- **#249** — `wasMissionActive` captured at L2478 BEFORE `Object.assign` at L2479; verified marker shape (L103, L2304) means missionActive IS overwritten, so pre-merge capture is necessary. `makeHaloGraphics()` byte-identical to lifted inline code (same Graphics+circle(0,0,16)+stroke). z-order `addChildAt(_, 0)` matches original create-time index. Idle→active branch is mutually exclusive with the dead-transition branch — no double-fire. Active→idle case handled via `else if (existing.halo)` visible-toggle. Clean.
- **#250** — Listener registered AFTER `appRef.current = app` (L1481, sync) so first `change` fire is guaranteed to see a truthy `currentApp`. Cleanup (L1947-1951) runs BEFORE `appRef.current = null` (L2037) — no teardown race. `!winterSnowRef.current` guards against double-create on repeat OFF→ON→OFF cycles. Snapshot-driven restoration via `snapshotRef.current.winterActive` correctly resyncs state on lazy-build path; the dedicated useEffect at L4346 handles subsequent transitions.
- **#251** — Filter destroy at L1984-1998 runs BEFORE `app.destroy(true)` at L2043 (correct order; reversed would noop). `try/catch` body only logs in DEV and otherwise drops — acceptable for teardown path that immediately nulls `glow.filters`. Verified BlurFilter is the only filter on bandit glows (campGlow is plain Graphics).

**LOW (pre-existing, NOT a blocker, NOT introduced by this PR):** `dayNightFilter` (`new ColorMatrixFilter()` at L1592) has the same FBO-leak pattern — nulled at L2020 without `.destroy()`. Out of scope for v2.5.2; worth a follow-up issue.

Build: `pnpm --filter @clan-world/web build` GREEN — tsc -b + vite, 2060 modules, 6.74s.

#claude-subagent

PR #244, opened 2026-05-12
fix(cockpit + map): show dead clansmen as DEAD, rotated + darkened

claude-do:
## Summary

When all 4 clans died during today's demo, the cockpit kept showing clansmen as **Idle / ready** and the world map kept rendering their sprites in a standing/upright pose. The UI completely masked the wipe.

This PR makes dead clansmen visibly dead in three surfaces:

### 1. Cockpit (web) — `apps/web/src/components/cockpit/tabs/ClansmanTab.tsx`
- New `isDead?: boolean` field on `ClansmanRow`.
- When dead: mission label becomes **DEAD** (red, letter-spaced, the ID gets struck through), the ETA/cooldown column collapses to an em dash, the row drops to ~55% opacity with a darker background and danger border, and the hunger bar stops flagging starvation (a corpse is beyond hunger).
- `data-dead="true|false"` attribute added for E2E hooks.

### 2. Cockpit (android) — `apps/clan-world-mobile/.../ClansmanTab.kt`
- Same affordances ported to Compose: dimmed background + alpha, red DEAD label with line-through on the ID, em-dash status column, no starvation flag.
- `ClansmanWire` + `StubData.ClansmanRow` gain an `isDead` field (default `false` for back-compat with older payloads).

### 3. World map (Pixi) — `apps/web/src/WorldMap.tsx`
- `LiveClansmanMarker` gains `isDead`, plus direct refs to `body` (the sprite) and `halo` (active-mission ring).
- New `applyDeadVisualState(body, isDead)` helper applies `rotation = π/2` (90° lying down) and `tint = 0x808080` (mid-grey multiply, collapses dynamic range without losing silhouette). Anchor swaps from feet (0.5, 0.82) → center (0.5, 0.5) so the rotation pivots on the body, not on a vanished foot.
- Sync loop detects alive→dead transitions and re-poses the body in place (no node rebuild). Halo hides and any in-flight travel route gets torn down — corpses don't walk.
- Position loop short-circuits for dead markers: no bob animation, no carry indicator, no status text, no route line. Frozen at last-known wander point.
- Dead state is sourced from chain `ClansmanFullView.clansman.clansman.state == ClansmanState.DEAD` (enum=3, mirrors `IClanWorld.sol`).

### 4. Convex query — `apps/server/convex/clansmen.ts`
- `getClanClansmen` now reads `clansman.state` and exposes `isDead` on the returned `ClansmanRow`.
[...]

PR #262, opened 2026-05-12
fix(map): focus-mode dims other-clan clansmen (only selected clan stays bright)

claude-do:
## Summary

Fix focus-mode regression in the PixiJS world map: clicking a clan base correctly dims the rest of the map but **all clansmen** (every clan) were staying bright, defeating the focus visual.

Per Liam directive 2026-05-12, only the **selected** clan's clansmen should stay bright; clansmen from other clans should dim along with the scrim.

## Root cause

`applyClanFocus()` correctly sets `marker.node.alpha = 0.18` for other-clan live clansmen on selection — but the per-frame `updateLiveClansmanPositions()` loop unconditionally re-set `marker.node.alpha = 1` every tick (both for alive and dead clansmen), clobbering the dim back to full.

## Fix

`apps/web/src/WorldMap.tsx` — `updateLiveClansmanPositions()`:
- Compute `focusAlpha` per marker (0.18 if focus is active and marker is from another clan; else 1).
- Replace the two unconditional `marker.node.alpha = 1` lines (alive + dead branches) with `marker.node.alpha = focusAlpha`.
- Also re-apply `marker.route.line.alpha = isOtherClan ? 0.12 : 1` per-frame so routes created mid-focus (after selection) get dimmed too.

Same dim constants as `applyClanFocus()` (0.18 body, 0.12 route line) — single source of truth via shared logic, no new constants.

## Before / after

- **Before:** click base → terrain + zones + bases + other-clan-bases dim; ALL clansmen stay bright (selected clan's clansmen indistinguishable from others'.)
- **After:** click base → terrain + zones + bases + other-clan-bases dim; **only selected clan's clansmen stay bright**; other-clan clansmen dim to 0.18; their route lines dim to 0.12. Click empty → clears focus → everyone full-bright.

## Dead-clansman interaction (PR #244)

Dead clansmen have `body.alpha = 0.9` from `applyDeadVisualState`. Setting the container `marker.node.alpha = 0.18` multiplies onto the body alpha → effective 0.9 × 0.18 ≈ 0.16 for dead-other-clan during focus. Reads as "even more faded" vs alive-other-clan (just 0.18), which is the consistent behavior Liam requested.

## Test plan

- [ ] `pnpm --filter @clan-world/web build` — green (verified locally)
- [ ] Click a clan base on the map → map dims, only that clan's bases + clansmen stay bright; other clans' clansmen and route lines dim
- [ ] Click empty terrain → focus clears, all clansmen return to full alpha
- [ ] Dead clansmen from non-selected clan during focus appear noticeably more faded than alive other-clan clansmen

### DIFF
--- a/apps/web/src/WorldMap.tsx
+++ b/apps/web/src/WorldMap.tsx
@@ -2834,7 +2834,14 @@ export function WorldMap() {
     const drawn = drawnRef.current;
     if (drawn.liveClansmen.length === 0) return;
     const tickFloat = currentTickFloat();
+    // Focus mode (base-click): selected clan stays bright; other clans dim.
+    // Same constants as applyClanFocus() so the per-frame value matches the
+    // one-shot value set on selection — prevents the per-frame loop from
+    // clobbering the dim back to 1 (Liam 2026-05-12).
+    const focusClanId = selectedClanIdRef.current;
     for (const marker of drawn.liveClansmen) {
+      const isOtherClan = !!focusClanId && marker.clanId !== focusClanId;
+      const focusAlpha = isOtherClan ? 0.18 : 1;
       const from = regionWanderPoint(marker.regionKey, marker, 0);
       if (!from) continue;
       // Dead clansmen freeze at their last-known wander point — no bob, no
@@ -2844,7 +2851,11 @@ export function WorldMap() {
         marker.node.x = from.x;
         marker.node.y = from.y;
         marker.node.zIndex = Math.round(marker.node.y + 8);
-        marker.node.alpha = 1; // container alpha stays 1; body's own alpha dims the sprite
+        // Container alpha is normally 1 (body's own alpha dims the sprite).
+        // During focus mode, dead-OTHER-clan clansmen also dim — the focus
+        // dim multiplies onto the existing dead-body alpha (PR #244) so
+        // dead+other-clan reads as even more faded than alive+other-clan.
+        marker.node.alpha = focusAlpha;
         marker.node.scale.set(Math.max(0.95, layoutRef.current.scale * 1.08));
         marker.carry.label.text = '';
         marker.carry.targetFill = 0;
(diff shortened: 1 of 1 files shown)


claude-do:
**Claude subagent (Tier 1)** · round 1 · head `14affc6`

Verdict: **CLEAN**

- `selectedClanIdRef.current` is canonically set in two writers only (useEffect L2108, clearSelection L1382); ticker reads same ref — no skew.
- Null-focus state: `!!focusClanId &&` short-circuits → `isOtherClan = false` for every marker → all bright. Verified.
- Dead-alpha multiplication is correct + no double-application. `body.alpha = 0.9` (applyDeadVisualState) is the ONLY body writer; new code writes `marker.node.alpha` (container). Pixi multiplies → dead+other-clan reads as 0.9 × 0.18 = 0.162. Matches intent.
- Route line dim resets to 1 every frame when un-focused — actually fixes a latent stale-dim bug for routes created mid-focus.
- halo/carry/statusBg/statusText all `container.addChild`'d in `makeLiveClansmanMarker` L2344-2398 → container alpha multiplies onto them.
- Travel-tickers loop L1741 operates on `travelsRef.current[*].gfx`, a different collection from `drawn.liveClansmen[*].node` — no overlap, no double-write. Was already focus-aware.
- No AnimatedSprite/body-alpha bypass — `body.alpha` has exactly 2 writers (`applyDeadVisualState` 0.9 / `applyAliveVisualState` 1.0). The other `sprite.alpha` writes are on bandit/base/terrain sprites, not clansmen bodies.
- Constants parity with `applyClanFocus` (0.18 body, 0.12 route) — author's comment L2838-2842 documents the coupling.
- `pnpm tsc --noEmit` ✓ · `pnpm build` ✓

Optional (DEFER, not blocking): extract `FOCUS_DIM_BODY = 0.18` + `FOCUS_DIM_ROUTE = 0.12` to module consts so all 4 call sites share one source.

#claude-subagent

PR #260, opened 2026-05-12
feat(web): new world map (960x1280) + region polygon overlay + first-pass resize

claude-do:
## Summary

Replaces the world-map background with Liam's new 960x1280 image and lays the groundwork for visually tuning the region polygons against the new art.

Three coordinated changes in one PR:

1. **Map asset swap** — `apps/web/src/assets/world-map.png` replaced (814x1448 -> 960x1280, JPEG-source converted to PNG via PIL).
2. **`MAP_WIDTH` / `MAP_HEIGHT` constants** updated; `REF_W` / `REF_H` and the projX/projY projection auto-propagate.
3. **`SHOW_REGION_POLYGONS` debug overlay** (default `true`): renders each `REGIONS[*].polygon` as a 30% alpha fill + 85% alpha stroke on the `terrainBackground` layer (above the bg image, below the clan-zone halos and sprites). Lets you visually inspect + iterate the polygon coords against the new map.
4. **First-pass region polygon resize** for the new dimensions per directive ("farms / forest / mountains wider, rest roughly same size").

## Polygon resize strategy

All polygons scaled by `sx = 960/814 ≈ 1.179` and `sy = 1280/1448 ≈ 0.884`. Then the four "wider" regions get an additional inward-edge expansion of +30 to +45 px so they read as visibly larger on the new map.

| Region        | Old polygon                                                    | New polygon (after scale + inward widen)                          | Inward widening |
|---------------|----------------------------------------------------------------|--------------------------------------------------------------------|-----------------|
| forest        | `[[0,0],[400,0],[475,165],[420,355],[285,500],[0,555]]`        | `[[0,0],[512,0],[600,146],[525,314],[366,442],[0,491]]`            | +40 east        |
| mountains     | `[[545,0],[814,0],[814,510],[660,535],[525,420],[505,255],[485,135]]` | `[[603,0],[960,0],[960,451],[778,473],[579,371],[556,225],[532,119]]` | -40 west        |
| unicorn-town  | `[[400,375],[510,375],[610,500],[535,585],[410,585],[355,520]]` | `[[472,331],[601,331],[719,442],[631,517],[484,517],[419,460]]`    | (scale only)    |
| west-farms    | `[[0,555],[330,520],[425,760],[300,960],[0,985]]`              | `[[0,491],[434,460],[546,672],[394,849],[0,871]]`                  | +45 east        |
[...]

PR #243, opened 2026-05-12
feat(web): winter snowfall particle overlay

claude-do:
## Summary

Add a PixiJS snowfall particle overlay that activates when `worldSnapshot.winterActive` is true. ~100 pooled sprites fall from the top of the viewport with gentle sine-driven horizontal drift, fade in over ~400ms on winter onset, and fade out over ~1s when the season changes.

Reported by Liam after demo session 2026-05-11.

## Design

- **Source of truth:** `worldSnapshot.winterActive` (same field `TopHud.tsx` already reads from the `getSnapshot` query — currently surfaced via `any` cast pending the broader snapshot type).
- **Layer:** added directly to `app.stage` as a sibling of the `pixi-viewport`, so particles render in screen space — they fall straight down regardless of pan/zoom, sit above all in-world content (terrain, regions, sigils, bubbles), and naturally sit below the React DOM UI (TopHud, EventTicker, status badges).
- **Hand-rolled** (no new dep) — pre-allocated pool of `Sprite` instances sharing one generated 4×4 white-disc texture (`renderer.generateTexture({ target: Graphics.circle().fill() })` at `resolution: 2` for hi-dpi crispness). Per-frame work is pure position math; recycle when `y > screenH + 10` — no GC churn.
- **Drift:** `x = baseX + sin(t·freq + seed) · amp` with per-particle `seed`/`freq`/`amp` so flakes don't sway in lockstep.
- **Density:** 100 particles at default. Tunable via `createWinterSnow(app, { particleCount })`. Erred toward subtle — large rates feel like a blizzard on the small viewport.
- **Lifecycle:**
  - Created once after Pixi init, alongside the existing ticker callbacks.
  - `setActive(true)` on winter start → container becomes visible, alpha ramps 0→1 over ~400ms.
  - `setActive(false)` on winter end → alpha ramps 1→0 over ~1000ms, then `idle` (ticker callback early-returns).
  - `destroy()` in cleanup unregisters the ticker callback and disposes the shared texture so Pixi `app.destroy()` doesn't leak GPU memory.
- **`prefers-reduced-motion`:** detected at init via `window.matchMedia`; when set, the snow system is never created — no ticker overhead.

## Files changed

- `apps/web/src/effects/winterSnow.ts` (new) — self-contained particle system + `WinterSnowHandle` interface.
- `apps/web/src/WorldMap.tsx` — import, ref, create-on-init, cleanup, and a `useEffect` that mirrors `winterActive` into `snow.setActive()`.

## Build

[...]

PR #263, opened 2026-05-12
feat(map): winter map overlay with fade-in/out

claude-do:
## Summary

Adds a winter-themed map overlay sprite layered above the base world map, alpha-modulated based on `snapshot.winterActive`. Pairs with the existing #243 snow particle effect — both fade in together at the start of Winter and fade out together as the season ends.

Built on top of the new 1086x1448 base map from PR #260 (now merged into `dev`). Winter PNG matches base-map dimensions exactly so no re-projection is needed.

- New asset: `apps/web/src/assets/world-map-winter.png` (1086x1448, 3.8MB)
- New Pixi `Sprite` in `terrainBackground`, added immediately after the base map sprite, before region polygons / clan zones / clansmen (so gameplay overlays still render on top of the winter ground)
- Wall-clock alpha state machine: `idle | fade-in | active | fade-out`, driven by per-frame ticker callback + a snapshot-edge-detector `useEffect`
- Mid-fade reversals are seamless — start-alpha is captured at each transition (no blink if Winter ends mid-fade-in)
- Initial-state guard: if the world boots already in Winter, sprite snaps to `alpha=1` instead of slow-revealing existing state

## Tunable constants (top of `WorldMap.tsx`)

```ts
const WINTER_FADE_IN_MS = 1500;   // freeze — feels cinematic at 1s tick cadence
const WINTER_FADE_OUT_MS = 2000;  // spring thaw is intentionally slower
```

Not coupled to the heartbeat interval — fade still looks natural if the on-chain tick rate changes.

## Layer order in `terrainBackground`

1. `backgroundHitArea` (invisible pointer target)
2. `bgSprite` — base map (always alpha=1)
3. **`winterSpriteRef` — winter overlay (alpha 0..1)** ← NEW
4. Region polygons / labels (added in `buildScene`)
5. Clan zones live in `terrainAccents` (parent sibling) — render above

## Sync with snow particles (`winterSnow.ts`)

Both the snow particle system and this new sprite watch the same `snapshot.winterActive` boolean via separate `useEffect`s. They start their fades at the same render and finish independently — the particle effect uses its own internal 400ms fade-in / 1000ms fade-out, while the map overlay uses the constants above (slower, more cinematic, since the ground recoloring is the bigger visual moment).

## Cleanup

Pixi-init `useEffect`'s cleanup now unregisters the winter-fade ticker callback and destroys the sprite. Texture is left in the Pixi `Assets` cache (matches the base-map pattern) so a hot-reload mount gets a cache hit.

## Build / typecheck

- `pnpm --filter @clan-world/web build` — GREEN
- No new TypeScript errors
- Dist includes `world-map-winter-*.png` (3.8MB) alongside the existing `world-map-*.png`

## Visual UAT — required

The fade is only verifiable by watching a live Winter season transition. Two options:

- [ ] Fresh world reset, advance ticks until Winter begins → confirm: snow starts falling AND ground fades to snow-covered version over ~1.5s
- [ ] Catch a live game mid-Winter → confirm ground reads as snowy + holding at full opacity
- [ ] Watch the Winter → next-season transition → confirm: ground fades back to normal map over ~2s while snow particles also fade
- [ ] Reduced-motion test: enable `prefers-reduced-motion`, trigger Winter → particles should NOT spawn (existing behavior preserved); map overlay STILL fades (it's not motion in the same sense; can revisit if Liam wants both gated)

## Open questions

- Should the map overlay also respect `prefers-reduced-motion` and snap (instead of fade) in that mode? Currently it always fades. Easy to gate behind the same MQ check the snow particles use if Liam prefers — left out of scope for the v1.


claude-do:
**Claude subagent (Tier 1)** · round 1 · head `d89487e`

Verdict: **CLEAN**

Validated:
- **Z-order** correct: winter sprite added to `terrainBackground` AFTER base map (L1697) and BEFORE region polygons/labels (added later in `buildScene`, L1738+). `eventMode='none'` keeps pointer events flowing to overlays above.
- **Edge detection** drives off `snapshot.winterActive: boolean` — confirmed in `getSnapshot.ts:26` and `schema.ts:13`, same field used by `TopHud` and existing `winterSnow`.
- **State machine** math symmetric in both directions; triple-reversal (on→off mid-fade-in→on mid-fade-out) captures `sprite.alpha` at each edge, no blink.
- **Initial-state guard** works: `prev === null` branch snaps `alpha=1` on boot-in-Winter, no slow reveal.
- **Cleanup** correct: ticker callback removed, `sprite.destroy()` no-args (sprite only, leaves Assets-cached texture), refs nulled, `prevWinterActiveRef` reset to `null` for remount.
- **Snow coordination** non-disruptive — both effects read same `winterActive`; snow's 400ms-in/1000ms-out and map's 1500ms-in/2000ms-out are intentionally decoupled (snow settles fast, ground recolors slower for cinematic feel).
- **Pixi v8** APIs all correct (`Sprite.destroy()` default `{texture:false}`, `app.ticker.add/remove` symmetric ref).
- **Build** green (`pnpm --filter @clan-world/web build` → 6.02s). PIL-verified asset dims: 1086×1448 for both PNGs.

Non-blocking notes (not for this PR):
1. Bundle cost: 7.2MB terrain assets total. Lazy-load follow-up suggested.
2. Cleanup comment at L2106-2108 calls `{ texture: true }` a Pixi-v8 error — it isn't (v8 accepts it). Author's intent (don't blow away Assets cache) is correct; framing nit only.

#claude-subagent

PR #269, opened 2026-05-14
feat(web): snapshot localStorage cache + grace period on "no chain data"

claude-do:
## Problem

iOS Safari (and any browser that aggressively pauses background tabs) leaves the Convex `useQuery(api.getSnapshot.getSnapshot)` hook returning `undefined` for a beat while the websocket reconnects when the user returns to the PWA. That gap surfaces as a white *"no chain data yet"* flash over a bare terrain map — feels broken even though it's just a sub-second reconnect.

### Part A — `apps/web/src/hooks/useCachedSnapshot.ts` (new)

Generic hook that:
- Synchronously reads the last-known snapshot from `localStorage` in `useState`'s initializer (so the first render already has data — no `undefined` flash).
- Writes back to `localStorage` on every non-`undefined` `live` value.
- Cache key: `cw-snapshot-v1` — **bump on Convex schema migrations**.
- Hard cap: `MAX_CACHE_AGE_MS = 1 hour` — anything older is dropped, so we never surface wildly stale state.
- Tolerates `JSON.parse` failures, missing key, quota errors, private-mode localStorage rejections — all silent fallthrough to `undefined`.
- Generic over `T` so we don't pull the Convex API type into the hook (keeps it framework-agnostic + easy to unit-test later).

### Part B — `apps/web/src/WorldMap.tsx`

```ts
const liveSnapshot = useQuery(api.getSnapshot.getSnapshot);
const snapshot = useCachedSnapshot(liveSnapshot);
```

Plus a 5-second grace period before the `data-testid="no-chain-data-placeholder"` overlay can render:

```ts
const [showNoChainDataPlaceholder, setShowNoChainDataPlaceholder] = useState(false);
useEffect(() => {
  if (DEMO_MODE) return;
  const hasClans = !!snapshot?.clans && snapshot.clans.length > 0;
  if (hasClans) { setShowNoChainDataPlaceholder(false); return; }
  const timer = window.setTimeout(() => setShowNoChainDataPlaceholder(true), 5000);
  return () => window.clearTimeout(timer);
}, [snapshot]);
```

JSX render condition swapped from `!snapshot?.clans || snapshot.clans.length === 0` to `showNoChainDataPlaceholder`. Inner JSX unchanged.

## Behavior summary

1. **First-ever load** (no cache, no live yet): nothing shown for ≤5s, then placeholder fades in.
2. **Returning PWA user** (cache hit, live still reconnecting): last-known world renders immediately; `live` arrives within ~1-2s; placeholder timer never fires.
3. **Long reconnect** (>5s, cache stale or missing): placeholder appears at 5s, hides instantly when `live` returns with clans.
4. **DEMO_MODE**: effect short-circuits, placeholder stays hidden as before.

## Cache invariants

- Key `cw-snapshot-v1` — bump on snapshot shape change.
- 1-hour hard cap on cache age (prevents "the world from yesterday" rendering).
- All localStorage access wrapped in try/catch — Safari private mode is a no-op, not a crash.

## Testing

- `pnpm --filter @clan-world/web typecheck` — passes.
- `pnpm --filter @clan-world/web build` — passes (5.9s, no new warnings).
- No unit-test convention in `apps/web` yet (only Playwright e2e), so per the implementation spec no test infra was introduced just for this hook. Hook is intentionally simple + self-contained — straightforward to add a vitest later when the test suite lands.

## Risk

Low. Two narrow additions:
- A new generic hook with no external deps beyond React.
- A single state + effect + render-condition swap on the placeholder overlay. The placeholder still renders the same JSX; only the *when* changed.

Worst-case regression is the placeholder taking 5s longer to appear in the no-chain-data steady state, which is desirable.

### DIFF
--- a/apps/web/src/WorldMap.tsx
+++ b/apps/web/src/WorldMap.tsx
@@ -1,5 +1,6 @@
 import { useEffect, useMemo, useRef, useState } from 'react';
 import { useSafeQuery as useQuery } from './hooks/useSafeQuery';
+import { useCachedSnapshot } from './hooks/useCachedSnapshot';
 import { Application, Assets, BlurFilter, ColorMatrixFilter, Container, Graphics, Rectangle, Sprite, Text } from 'pixi.js';
 import { Viewport } from 'pixi-viewport';
 import { useAgentLogs, type AgentLog } from './useAgentLogs';
@@ -1322,7 +1323,15 @@ export function WorldMap() {
   const [selectedClanId, setSelectedClanId] = useState<string | null>(null);
 
   const logs = useAgentLogs();
-  const snapshot = useQuery(api.getSnapshot.getSnapshot);
+  const liveSnapshot = useQuery(api.getSnapshot.getSnapshot);
+  // Cache the snapshot in localStorage so iOS Safari (and other browsers that
+  // pause background tabs) can render the last-known world state instantly on
+  // PWA return — instead of flashing the "no chain data yet" placeholder
+  // while Convex's websocket reconnects.
+  const snapshot = useCachedSnapshot(liveSnapshot);
+  // Grace period before showing the "no chain data yet" placeholder — see the
+  // useEffect a few hooks below for the 5s debounce.
+  const [showNoChainDataPlaceholder, setShowNoChainDataPlaceholder] = useState(false);
   const rawChainEvents = useQuery(api.events.getRecentChainEvents) as ChainEvent[] | undefined;
 
   // Derived live tick counter — the worldSnapshot.tick field is currently
@@ -1371,6 +1380,23 @@ export function WorldMap() {
     if (typeof rawTick === 'number' && Number.isFinite(rawTick)) {
       tickClockRef.current = { tick: rawTick, seenAtMs: Date.now() };
     }
+  }, [snapshot]);
+
+  // 5-second grace period before showing the "no chain data yet" placeholder.
+  // If chain data is present, hide immediately. If it's empty, wait 5s before
+  // flipping the placeholder on — so fast Convex reconnects (e.g. iOS Safari
+  // PWA returns) never reveal the bare-terrain overlay. Combined with the
+  // localStorage cache in useCachedSnapshot above, most reconnects render the
+  // last-known state immediately and never trip this timer at all.
+  useEffect(() => {
+    if (DEMO_MODE) return;
+    const hasClans = !!snapshot?.clans && snapshot.clans.length > 0;
+    if (hasClans) {
+      setShowNoChainDataPlaceholder(false);
+      return;
+    }
+    const timer = window.setTimeout(() => setShowNoChainDataPlaceholder(true), 5000);
+    return () => window.clearTimeout(timer);
   }, [snapshot]);
 
   // Snapshot-diff: derive bandit resolution outcome the moment the chain transitions
@@ -5327,9 +5353,11 @@ export function WorldMap() {
       })()}
 
       {/* "No chain data yet" placeholder — shown when DEMO_MODE is off and no
-          live snapshot clans are present. Centered overlay so it reads clearly
-          on the bare terrain map. */}
-      {!DEMO_MODE && (!snapshot?.clans || snapshot.clans.length === 0) && (
+          snapshot clans are present, *after* a 5-second grace period managed
+          by `showNoChainDataPlaceholder` (see the useEffect declared with the
+          snapshot hooks above). Centered overlay so it reads clearly on the
+          bare terrain map. */}
+      {!DEMO_MODE && showNoChainDataPlaceholder && (
         <div
           data-testid="no-chain-data-placeholder"
           style={{
--- a/apps/web/src/WorldMap.tsx
+++ b/apps/web/src/WorldMap.tsx
@@ -1329,6 +1329,10 @@ export function WorldMap() {
   // PWA return — instead of flashing the "no chain data yet" placeholder
   // while Convex's websocket reconnects.
   const snapshot = useCachedSnapshot(liveSnapshot);
+  // Stable boolean: true only when the snapshot contains at least one deployed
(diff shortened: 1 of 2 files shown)


claude-do:
**Codex review** · round 1 · head `6d6cbd9`

Verdict: NEEDS_FIXES (0 HIGH, 1 MEDIUM, 0 LOW)

- [MED] apps/web/src/hooks/useCachedSnapshot.ts:18 — `CACHE_KEY = 'cw-snapshot-v1'` is origin-scoped; if dev and prod share the same origin with path-based routing, dev writes can overwrite prod's cached snapshot. Include a `VITE_CONVEX_URL` or `VITE_APP_ENV` discriminator in the key.

All other concerns (SSR safety, redundant re-render, grace-period cleanup, JSON type safety, private-mode rejection) are CLEAN.

#codex

claude-do:
**Claude subagent (orch tier 1)** · round 1 · head `6d6cbd9`

Verdict: NEEDS_FIXES (1 HIGH, 2 MED)

- [HIGH] `apps/web/src/hooks/useCachedSnapshot.ts` — `CACHE_KEY` (`cw-snapshot-v1`) has no env suffix; cross-environment cache pollution on shared origins (e.g. localhost dev vs staging, Vercel preview deployments). Fix: append hash of `import.meta.env.VITE_CONVEX_URL` to key.
- [MED] `apps/web/src/hooks/useCachedSnapshot.ts` — Cached payload shape not validated; a schema change without manual `CACHE_KEY` bump silently serves structurally-wrong data typed as `T`. Add a version discriminant field validated on read.
- [MED] `apps/web/src/hooks/useCachedSnapshot.ts` — `setCached(live)` fires on every Convex tick even when `cached` is irrelevant to the return value (`live ?? cached = live`). Causes a redundant re-render in the large WorldMap component per poll cycle. Guard with `if (live !== cachedRef.current)`.

Grace-period timer cleanup, mount-time race, and localStorage quota handling are all correct.

#claude-subagent

claude-do:
**Gemini flash review** · round 1 · head `6d6cbd9`

Verdict: CLEAN

No bugs, null-check gaps, hook violations, or type errors found. One low candidate (DEMO_MODE missing from useEffect dep array in WorldMap.tsx) verified false positive — it is a build-time static constant that never changes at runtime.

#gemini-flash

claude-do:
## Fix round 2 — addressed local swarm feedback

**Commits:** \`a3db4b4\` fix(web): scope snapshot cache key per env + version payload shape (PR #269 r2)

### Addressed

- **[Tier 1] apps/web/src/hooks/useCachedSnapshot.ts:18** — cache key not scoped per environment; dev + prod share same localStorage slot when running on localhost → fixed in \`a3db4b4\` (append \`VITE_CONVEX_URL\` to \`CACHE_KEY\` via \`ENV_SCOPE\`, fallback to \`'no-backend'\` when unset)
- **[Tier 1 + Tier 2] apps/web/src/hooks/useCachedSnapshot.ts:22** — cached payload shape not versioned; future Convex schema changes could serve corrupt state without an explicit invalidation path → fixed in \`a3db4b4\` (added \`PAYLOAD_VERSION = 1\` discriminant to \`Cached<T>\`, validated on read, written on write)

### Skipped / false positive

- **[Tier 1] apps/web/src/hooks/useCachedSnapshot.ts:46** — \`setCached(live)\` flagged as redundant re-render. Skipped: the cached state is intentionally retained so future reconnect windows where \`live\` goes back to \`undefined\` can still surface the cached value. Tier 2 (codex) correctly identified this as intentional; cross-tier dispute resolved in favor of keep.

### Local state after fix

- Type-check: pass
- Build: pass (vite 5.4.11, 2070 modules, 5.79s)
- Local swarm: Codex CLEAN, Gemini flash UNAVAILABLE (hung without output, per known pattern), Claude self-review CLEAN

### Not re-cycling cloud reviewers
Per cloud-thrift policy: local swarm is the iteration loop, cloud is a single final sanity pass.

claude-do:
## Fix round 3 — addressed gemini-code-assist cloud reviewer feedback

**Commits:** `498aadb` fix(web): narrow grace-period dep + throttle localStorage writes (PR #269 r3)

### Addressed

- **[gemini-code-assist] apps/web/src/WorldMap.tsx:~1400** — Grace-period timer reset on every heartbeat tick because `useEffect` depended on the whole `snapshot` object (which changes every tick). Fixed in `498aadb`: extracted `hasClans` boolean adjacent to the `snapshot` declaration; changed dep array from `[snapshot]` to `[hasClans]`. Timer now only resets when clan data-presence changes.

- **[gemini-code-assist] apps/web/src/hooks/useCachedSnapshot.ts:~55** — `localStorage.setItem` + `JSON.stringify` called on every Convex heartbeat tick, risking main-thread jank on mobile Safari. Fixed in `498aadb`: throttle to once per 30 seconds via `lastWriteTsRef`. First load still writes immediately (ref initializes to 0); in-memory `setCached(live)` still runs every tick so component state stays fresh.

### Local state after fix

- Type-check: pass (tsc --noEmit clean)
- Build: pass (Vite built in 5.91s)
- Local swarm: Codex CLEAN (`codex review --base origin/feat/snapshot-cache-and-grace-period`), Gemini flash CLEAN (explicit "CLEAN." verdict with substantive confirmation of both fixes), Claude tier-1 CLEAN (inline analysis)

### Not re-cycling cloud reviewers

Per cloud-thrift policy: local swarm is the iteration loop, cloud is a single final sanity pass.

claude-do:
**Round 3.1 — additional codex tail finding**

After round-3 landed, an out-of-band codex review flagged one residual:

- **[LOW]** `useCachedSnapshot.ts:57-65` — `lastWriteTsRef.current = now` only advanced after a successful `setItem`. In private-mode / quota-rejected paths the ref stayed at 0, so the throttle was bypassed in exactly the failure mode the catch was meant to tolerate. Fixed in `cdc918d` by moving the timestamp update BEFORE the try block.

Typecheck + build clean.

#claude-orch

PR #270, opened 2026-05-14
feat(web): static HTML ghost layer (eliminate PixiJS warmup flash)

claude-do:
## Summary

Mounts a static HTML+CSS \"ghost\" of the map UNDER the PixiJS canvas on cold load. Renders the cached map BG + clan bases as plain `` tags using the snapshot + viewport state already persisted by PR #269. Fades out (200ms) once PixiJS dispatches its existing `pixiReady` signal.

**Net effect on iOS Safari PWA return:** zero-millisecond perceived first paint instead of the ~500ms–1s WebGL warmup gap that PR #269 could not address (it fixed the data flash; this fixes the canvas flash).

## Approach

Implements **Option D** from the recent design audit. Two parallel DA reviews (codex + gemini-pro) both ranked D 1st out of 4 candidate approaches.

- New: `apps/web/src/components/MapGhostLayer.tsx`
- Modified: `apps/web/src/WorldMap.tsx`
  - Imports + renders `<MapGhostLayer pixiReady={pixiReady} />` inside the existing `canvasWrapRef` div.
  - Pins the Pixi canvas as `position: absolute; inset: 0; z-index: 1` so the absolutely-positioned ghost (z-index 0) sits beneath it in the same stacking context. Without explicit z-index on both children, the absolutely-positioned ghost would paint OVER the static-positioned canvas, defeating the swap.
  - Trivial pre-existing typecheck fix: `REGIONS[i].polygon` possibly-undefined guard at line 4058 (was failing `pnpm typecheck` + `pnpm build` on `dev` HEAD; needed green to verify this PR).

### Coordinate math

CSS transform mirrors `pixi-viewport.moveCenter(cx, cy)` + `setZoom(scale)`:

```css
transform-origin: 0 0;
transform: translate(Sw/2, Sh/2) scale(s) translate(-cx, -cy);
```

applied to a wrapper containing children in WORLD-pixel coords. Maps each child at `(wx, wy)` to screen `(Sw/2 + (wx-cx)*s, Sh/2 + (wy-cy)*s)` — identical to pixi-viewport's projection.

### Reads the same state the canvas reads

- `cw-snapshot-v1:${VITE_CONVEX_URL}` from localStorage (same key derivation as `useCachedSnapshot.ts`, same validation: v=1, ts within 1h, `clans[]` non-empty).
- `cw-viewport-v1` from sessionStorage. Falls back to fit-cover (matches Pixi's cold-init default) if absent.

### Static, additive, non-invasive

- Does NOT touch the PixiJS rendering pipeline.
- Does NOT re-subscribe to Convex (no per-tick re-renders).
- Does NOT modify `useCachedSnapshot.ts` or the grace-period logic.
- Reads cached clans ONCE per mount; the ghost is only on screen for ~500ms.

## Blind spots from DA — addressed

- **DPI** (gemini-pro): cached `(cx, cy, scale)` are in CSS-pixel space because pixi-viewport's API works in stage coords (not device). The HTML layer is also in CSS-pixel space. No conversion needed.
- **Asset path** (codex): map background imported as a Vite asset (`import worldMapBg from '../assets/world-map.png'`) → hashed URL → HTTP-cached with content hash. NOT moved to `public/`.
- **Coordinate sync** (both): the ghost uses the same `(cx, cy, scale)` triple that pixi-viewport persisted, so when Pixi takes over there's no visual snap. Fit-cover fallback matches Pixi's cold-init `initialFitScale`.
- **Stepwise base-size formula** (caught during impl): `baseVisualScale` in `WorldMap.tsx` is stepwise (1-2 → 0.675, 3-4 → 0.81, 5 → 0.9), not linear. Mirrored exactly in `MapGhostLayer.tsx`.

## Trade-offs / deferred

- **No `<link rel=\"preload\">` for the map background.** Vite hashes the asset URL so a static preload tag in `index.html` needs a build-time plugin (`vite-plugin-preload`) or a `transformIndexHtml` hook. Deferred — relying on browser HTTP cache is the cheapest path for v1. Filed as follow-up.
- **REGIONS / clan metadata duplicated** in `MapGhostLayer.tsx`. The alternative — importing from `WorldMap.tsx` — would pull in the PixiJS bundle through `WorldMap`'s top-level imports, defeating the \"render before canvas-bundle parses\" goal. The duplicates are flagged inline with \"keep in sync\" notes.

## Verification

- `pnpm install` — clean.
- `pnpm --filter @clan-world/web typecheck` — passes (was failing pre-existing TS2532 on `dev`, fixed inline).
- `pnpm --filter @clan-world/web build` — passes, no new chunks, no new bundle warnings.

## Test plan

- [ ] Hard refresh on `/`: ghost visible briefly (~hundreds of ms on a cold WebGL init), then fades as PixiJS takes over with no visual snap.
- [ ] Zoom in, refresh: ghost renders at the saved zoom (not fit-cover) so the swap to canvas is positionally invariant.
- [ ] iOS Safari PWA: background the tab for 30+ seconds, return — perceived first paint should be instantaneous (cached BG + bases visible) instead of black + delayed WebGL init.
- [ ] First-ever visit (clear storage): ghost shows only the map background (no bases). The PR #269 grace-period placeholder appears after 5s as before.

### DIFF
--- a/apps/web/src/WorldMap.tsx
+++ b/apps/web/src/WorldMap.tsx
@@ -7,6 +7,7 @@ import { useAgentLogs, type AgentLog } from './useAgentLogs';
 import { WorldNoticePanel } from './WorldNoticePanel';
 import { TopHud } from './TopHud';
 import { EventTicker } from './EventTicker';
+import { MapGhostLayer } from './components/MapGhostLayer';
 import { api } from '../../server/convex/_generated/api';
 import worldMapBg from './assets/world-map.png';
 import worldMapWinterBg from './assets/world-map-winter.png';
@@ -1645,8 +1646,15 @@ export function WorldMap() {
         if (!mounted || !isMountedRef.current || !canvasWrapRef.current) return;
         canvasWrapRef.current.appendChild(app.canvas);
         pixiCanvas = app.canvas;
-        // CSS: stretch to wrapper
+        // CSS: stretch to wrapper. We pin the canvas as `position: absolute;
+        // inset: 0` with an explicit z-index ABOVE the MapGhostLayer (which
+        // sits at z-index 0 in the same container). Without explicit
+        // positioning + z-index the absolutely-positioned ghost would paint
+        // on top of the static-positioned canvas, defeating the swap.
         app.canvas.style.display = 'block';
+        app.canvas.style.position = 'absolute';
+        app.canvas.style.inset = '0';
+        app.canvas.style.zIndex = '1';
         app.canvas.style.width = '100%';
         app.canvas.style.height = '100%';
         // Round-6 pinch fix: pixi-viewport handles multi-touch internally, so
@@ -4054,8 +4062,9 @@ export function WorldMap() {
       // One Text per polygon vertex, showing its (x,y) polygon coords. Added to
       // terrainBackground AFTER the Graphics so labels render on top of the fill.
       const vertexLabels: Text[] = [];
-      if (SHOW_REGION_POLYGONS) {
-        for (const [px, py] of REGIONS[i].polygon) {
+      const regionDef = REGIONS[i];
+      if (SHOW_REGION_POLYGONS && regionDef) {
+        for (const [px, py] of regionDef.polygon) {
           const t = new Text({
             text: `(${px},${py})`,
             style: {
@@ -5349,7 +5358,15 @@ export function WorldMap() {
           width: '100%',
           height: '100%',
         }}
-      />
+      >
+        {/* Static HTML "ghost" of the map — visible briefly while PixiJS
+            warms up WebGL (~500ms-1s on cold mobile Safari). Renders the
+            cached map background + clan bases using the same viewport
+            (cx, cy, scale) state pixi-viewport already persists. Fades
+            out once `pixiReady` flips true.
+            See: apps/web/src/components/MapGhostLayer.tsx. */}
+        <MapGhostLayer pixiReady={pixiReady} />
+      </div>
 
       {/* Top HUD bar — tick clock, season progress, status chips */}
       <TopHud liveTick={liveTick} />
--- a/apps/web/src/WorldMap.tsx
+++ b/apps/web/src/WorldMap.tsx
@@ -1646,11 +1646,11 @@ export function WorldMap() {
         if (!mounted || !isMountedRef.current || !canvasWrapRef.current) return;
         canvasWrapRef.current.appendChild(app.canvas);
         pixiCanvas = app.canvas;
-        // CSS: stretch to wrapper. We pin the canvas as `position: absolute;
-        // inset: 0` with an explicit z-index ABOVE the MapGhostLayer (which
-        // sits at z-index 0 in the same container). Without explicit
-        // positioning + z-index the absolutely-positioned ghost would paint
-        // on top of the static-positioned canvas, defeating the swap.
+        // CSS: stretch to wrapper. The ghost (MapGhostLayer) sits at z-index 2
+        // so it covers the canvas during warmup; the canvas sits at z-index 1
+        // BELOW the ghost. Once `pixiReady` fires the ghost fades out and
+        // unmounts, leaving only the canvas. pointer-events:none on the ghost
+        // ensures taps reach the canvas even before the fade completes.
         app.canvas.style.display = 'block';
         app.canvas.style.position = 'absolute';
         app.canvas.style.inset = '0';
--- a/apps/web/src/WorldMap.tsx
+++ b/apps/web/src/WorldMap.tsx
@@ -8,6 +8,7 @@ import { WorldNoticePanel } from './WorldNoticePanel';
 import { TopHud } from './TopHud';
 import { EventTicker } from './EventTicker';
 import { MapGhostLayer } from './components/MapGhostLayer';
+import { MAP_WIDTH, MAP_HEIGHT, LIVE_CLAN_REGION_BY_ID } from './components/mapGeometry';
 import { api } from '../../server/convex/_generated/api';
 import worldMapBg from './assets/world-map.png';
 import worldMapWinterBg from './assets/world-map-winter.png';
@@ -29,8 +30,7 @@ import {
 // World dimensions used by pixi-viewport for pan/clamp/center math.
 // Matches the actual hand-curated bg PNG (apps/web/src/assets/world-map.png)
 // at native resolution. The viewport scales/pans this world inside the screen.
-const MAP_WIDTH = 1086;
-const MAP_HEIGHT = 1448;
+// Re-exported from mapGeometry below (pure-data shared module).
 const WORLD_WIDTH = MAP_WIDTH;
 const WORLD_HEIGHT = MAP_HEIGHT;
 
@@ -319,14 +319,7 @@ const MOCK_CLANS: ClanDef[] = [
   { id: 'clan-storm', spriteSheetId: 'clan-storm', name: 'Storm Riders', homeRegion: 'east-farms', color: 0x44aacc, sigil: '/sigils/storm-riders-sigil.png', portrait: '/portraits/mira-portrait.png',   archetype: 'Trader',     basePng: '/bases/tide-wardens.png',  clansmanPng: '/clansmen/clan-storm.png' },
 ];
 
-const LIVE_CLAN_REGION_BY_ID: Record<number, string> = {
-  1: 'forest',
-  2: 'mountains',
-  4: 'west-farms',
-  5: 'east-farms',
-  6: 'west-docks',
-  7: 'east-docks',
-};
+// LIVE_CLAN_REGION_BY_ID is imported from ./components/mapGeometry — shared with MapGhostLayer.
 
 function clansmanPngForClanId(clanId: string): string | null {
   return MOCK_CLANS.find((clan) => clan.id === clanId)?.clansmanPng ?? null;
--- a/apps/web/src/WorldMap.tsx
+++ b/apps/web/src/WorldMap.tsx
@@ -5357,6 +5357,9 @@ export function WorldMap() {
           inset: 0,
           width: '100%',
           height: '100%',
+          // Scope the ghost/canvas z-index stacking to this container so it
+          // cannot compete with grand-sibling overlays (HUD, scoreboard, etc).
+          isolation: 'isolate',
         }}
       >
         {/* Static HTML "ghost" of the map — visible briefly while PixiJS
--- a/apps/web/src/components/MapGhostLayer.tsx
+++ b/apps/web/src/components/MapGhostLayer.tsx
@@ -1,9 +1,18 @@
 import { useEffect, useMemo, useRef, useState } from 'react';
 import worldMapBg from '../assets/world-map.png';
+import {
+  MAP_WIDTH,
+  MAP_HEIGHT,
+  REGION_CENTERS_BY_KEY,
+  LIVE_CLAN_REGION_BY_ID,
+  BASE_PNG_BY_VISUAL_INDEX,
+  FALLBACK_HOME_REGION_BY_INDEX,
+  baseVisualScale as _baseVisualScale,
+} from './mapGeometry';
 
 /**
- * MapGhostLayer — static HTML+CSS "ghost" of the world map rendered UNDER the
- * PixiJS canvas while WebGL is warming up.
+ * MapGhostLayer — static HTML+CSS "ghost" of the world map rendered ABOVE the
+ * PixiJS canvas while WebGL is warming up, then fading out once pixiReady fires.
  *
  * Why: on iOS Safari (and other browsers that aggressively pause background
  * tabs), returning to the PWA goes through this sequence:
@@ -35,8 +44,6 @@ import worldMapBg from '../assets/world-map.png';
  * pixi-viewport's projection.
  */
 
-const MAP_WIDTH = 1086;
-const MAP_HEIGHT = 1448;
 const VIEWPORT_STORAGE_KEY = 'cw-viewport-v1';
 // Matches the same key derivation used by useCachedSnapshot.ts so the ghost
 // reads the exact same payload that the canvas will hydrate from.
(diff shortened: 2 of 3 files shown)


claude-do:
**Codex review** · round 1 · head `a98417c`

Verdict: NEEDS_FIXES (0 HIGH, 1 MEDIUM, 1 LOW)

- [MED]  apps/web/src/WorldMap.tsx:1654 — `z-index:1` on canvas not scoped by a parent stacking context; canvas can occlude absolutely-positioned HUD/overlay siblings and intercept their touch events. Fix: give the canvas wrapper `isolation:isolate` or `position:relative; z-index:0` to create a containing stacking context.
- [LOW]  apps/web/src/components/MapGhostLayer.tsx:346 — `decoding="async"` on the map background `` means ghost can paint blank on cold cache, partially reintroducing the warmup flash this PR is meant to eliminate. Use `decoding="sync"` or `loading="eager"` + preload link for the main background image.

No leaks found (ResizeObserver disconnects, fade timer clears). Fade-out order is safe. Transform math not flagged. Constants duplication rationale correct (Vite tree-shaking alone won't prevent bundle bleed from a Pixi-importing module).

#codex

claude-do:
**Claude subagent (orch tier 1)** · round 1 · head `a98417c`

Verdict: CONDITIONAL APPROVE — one real bug (fade race), one dead-code line; everything else is sound.

- [MEDIUM] `apps/web/src/components/MapGhostLayer.tsx:332` — setTimeout delay equals CSS transition duration (both 200ms); on a busy mobile scheduler the unmount fires before the final opacity frame composites, producing a snap. Fix: delay unmount by FADE_OUT_MS + 50ms (or use `transitionend` listener).
- [LOW] `apps/web/src/components/MapGhostLayer.tsx:401` — `visibility: pixiReady && fullyHidden ? 'hidden' : 'visible'` is dead code; `return null` at line 336 always fires first when `fullyHidden` is true. Remove.
- [INFO] `apps/web/src/WorldMap.tsx:4062` — null guard on `REGIONS[i]` is a drive-by typecheck fix unrelated to this PR's scope. Confirm dev HEAD tsc state; if it was clean, extract to a separate 1-line fix PR.
- [INFO] `apps/web/src/components/MapGhostLayer.tsx:122` — Duplicate constants vs WorldMap.tsx are intentional: WorldMap has top-level PixiJS side effects that Vite cannot tree-shake, so importing from it would pull the full PixiJS bundle into the ghost chunk. Bundle concern is real. Consider a shared `world-constants.ts` extraction in a follow-up.
- Z-index layering, hit testing, transform math, DPI handling all verified correct (canvasWrapRef contains canvas z-index, overlays are grand-siblings in their own stacking context).

#claude-subagent

claude-do:
**Gemini flash review** · round 1 · head `a98417c`

Verdict: NEEDS_FIXES (1 HIGH, 1 MEDIUM, 3 LOW)

- [HIGH]  WorldMap.tsx:1654 / MapGhostLayer.tsx:348 — z-index reversal: canvas (z=1) covers ghost (z=0) immediately; ghost never visible during warmup. Set ghost > canvas + pointer-events: none
- [MED]   MapGhostLayer.tsx:327 — screenW/screenH init to 0; ResizeObserver fires post-paint, causing one-frame flash of misplaced map. Init from window.innerWidth/innerHeight
- [LOW]   MapGhostLayer.tsx:401 — imageRendering: 'auto' blurs pixel art; use 'pixelated' to match PixiJS Nearest filter
- [LOW]   WorldMap.tsx:4065 — regionDef null check present but regionDef.polygon not guarded before iteration
- [LOW]   MapGhostLayer.tsx:112 — LIVE_CLAN_REGION_BY_ID missing IDs 3 + 8 (unicorn-town, deep-sea); falls back to wrong region on ghost

#gemini-flash

claude-do:
## Round 2 — addressed local-swarm round-1 feedback (commit `7a8ed11`)

### Addressed (7 items)

**HIGH** (gemini-flash):
- Ghost z-index reversal — moved ghost from z=0 to z=2 (above canvas at z=1). PixiJS canvas was covering the ghost on its first opaque frame. Now ghost is visible from first paint until `pixiReady` fades it out. `pointer-events: none` ensures it never blocks taps.

**MED** (3):
- Fade-out timer race (tier 1) — bumped setTimeout from `FADE_OUT_MS` to `FADE_OUT_MS + 50` so the browser compositor delivers the final opacity=0 frame before unmount.
- ResizeObserver init delay (gemini-flash) — `dims` useState initializer now reads `window.innerWidth/innerHeight` synchronously instead of starting at 0. First paint uses real dimensions.
- Stacking context isolation (codex) — added `isolation: isolate` to the canvas-wrap div so the canvas + ghost stacking is scoped from grand-sibling overlays.

**LOW** (3):
- Map BG `decoding="sync"` (was `async`) — async decode partially reintroduced the flash this PR is meant to eliminate.
- `imageRendering: pixelated` (was `auto`) — matches PixiJS's Nearest filter; no visual change on swap.
- Removed dead `visibility:` line — `return null` at the `fullyHidden` check renders it unreachable.

### Skipped (false positives)

- `regionDef.polygon` not guarded — region.polygon is always defined; not undefined-able.
- `LIVE_CLAN_REGION_BY_ID` missing IDs 3/8 — those are region IDs (unicorn-town, deep-sea), not clan IDs. Only 4 clans exist (iron/ember/dawn/storm), all mapped.

### Kept (confirmed legit)

- The round-1 drive-by null guard at `WorldMap.tsx:4062` stays — verified that `origin/dev`'s WorldMap.tsx already fails tsc with TS2532 at line 4058 (`REGIONS[i]` is `T | undefined` under noUncheckedIndexedAccess). The guard makes typecheck pass.
- Duplicate constants between `MapGhostLayer.tsx` and `WorldMap.tsx` — tier 2 confirmed Vite cannot tree-shake the PixiJS bundle bleed; duplication is intentional bundle hygiene.

Build + typecheck clean.

#claude-orch

claude-do:
**Codex review** · round 2 · head `7a8ed11`

Verdict: NEEDS_FIXES (0 HIGH, 0 MEDIUM, 1 LOW)

- [LOW] apps/web/src/components/MapGhostLayer.tsx:237 — window-based dims init: ghost renders at wrong scale/center on first paint when container is smaller than viewport; one-frame visual jump before mount measurement corrects it. Fix: synchronous container measure via useLayoutEffect or delay transform-sensitive render until rootRef has real dimensions.

z-index fix (ghost z=2, canvas z=1, pointer-events:none) is coherent. isolation:isolate scopes the stacking context correctly. FADE_OUT_MS+50 timer cleanup is race-safe. decoding=sync is a perf trade-off, not a correctness bug.

#codex

claude-do:
**Claude subagent (orch tier 1)** · round 2 · head `7a8ed11`

Verdict: CLEAN — all 7 round-1 findings correctly addressed. Ready to merge.

- **Z-index fix (ghost z=2, canvas z=1, pointer-events:none)**: CORRECT. Stack is sound within `isolation:isolate` scope; HUD/scoreboard siblings unaffected.
- **Fade +50ms headroom**: CLEAN. No new race introduced — extends safe unmount window only.
- **window.innerWidth dims init**: LOW / KNOWN ARTIFACT. First paint over-renders ghost at window size if map occupies a sub-region. Clipped by `overflow:hidden` + blocked by `pointer-events:none` — no functional breakage. useEffect refines to `el.clientWidth` post-mount. Not a regression vs. prior 0×0 blank. Codex r2 flagged the same; not a merge blocker.
- **isolation:isolate on canvasWrapRef**: CLEAN. Scopes z-index contest to the wrapper.
- **decoding="sync"**: ACCEPTABLE TRADEOFF. Sync decode prevents a blank-ghost frame on slow-decode devices. Correct call.
- **imageRendering: pixelated (map BG only)**: CORRECT. Matches PixiJS Nearest filter. Base sprites intentionally untouched.
- **Dead-code visibility removal**: CLEAN. `fullyHidden → return null` already removes from DOM and a11y tree.

#claude-subagent

claude-do:
## Fix round 2 — addressed reviewer feedback

**Commits:** `7a8ed11` initial r2 fixes + `6e2d277` hooks-order critical fix + comment fix

### Addressed

- **[gemini-flash] MapGhostLayer.tsx:~329** — ghost z-index was 0, canvas was 1; PixiJS canvas covered ghost on first opaque frame, defeating the warmup-eliminator. Fixed: ghost now `zIndex: 2`, canvas stays at `zIndex: 1`. Ghost has `pointer-events: none` so taps reach canvas during fade.
- **[tier 1 / Claude] MapGhostLayer.tsx:~268** — fade-out `setTimeout` matched CSS transition duration exactly (both 200ms); on slow mobile scheduler, unmount could fire before final opacity frame composites → 1-frame snap. Fixed: bump to `FADE_OUT_MS + 50` (250ms total).
- **[tier 3 / gemini-flash] MapGhostLayer.tsx:~237** — `screenW`/`screenH` useState initialized to `null`; ResizeObserver fires post-paint so first render saw 0×0 dimensions → ghost rendered at origin or invisible. Fixed: init synchronously from `window.innerWidth`/`window.innerHeight`.
- **[tier 2 / codex] WorldMap.tsx:~5353** — canvas wrap div had no stacking context isolation; z-indices could compete with grand-sibling overlays. Fixed: added `isolation: 'isolate'` to the `canvasWrapRef` div.
- **[tier 2 / codex] MapGhostLayer.tsx:~362** — map background `` deferred decode, partially reintroducing the blank flash this PR was meant to eliminate. Fixed: `decoding="sync"`.
- **[tier 3 / gemini-flash] MapGhostLayer.tsx:~373** — `imageRendering: 'auto'` (bilinear) caused a visual change when canvas took over (PixiJS uses Nearest filter). Fixed: `imageRendering: 'pixelated'`.
- **[tier 1 / Claude] MapGhostLayer.tsx:~337** — dead-code `visibility` line: when `fullyHidden` is true, `return null` at line 272 already fires before that line is reached. Removed.
- **[codex sub-round] MapGhostLayer.tsx:~280** — **CRITICAL (Rules of Hooks):** `useMemo` for `viewport` was positioned after `if (fullyHidden) return null`. When `fullyHidden` becomes true React sees fewer hooks than expected and throws `Rendered fewer hooks than expected`. Fixed: moved `viewport` useMemo above the early return.
- **[gemini-flash sub-round] WorldMap.tsx:~1649** — comment said ghost is at z-index 0 (stale from r1 original design). Fixed: updated comment to reflect ghost=2, canvas=1.

### Skipped / false positive

- **[tier 3] MapGhostLayer.tsx** — `regionDef.polygon` not guarded. False positive: `region` lookup already returns `null` for any undefined key and we bail with `if (!region) return null`.
- **[tier 3] MapGhostLayer.tsx:72** — `LIVE_CLAN_REGION_BY_ID` missing IDs 3 and 8. False positive: IDs 3 (unicorn-town) and 8 (deep-sea) are region slots with no live clan assigned. Only 4 clans map to IDs 1/2/4/5/6/7.
- **[tier 1] MapGhostLayer.tsx** — duplicate constants. Intentional: ghost layer cannot import WorldMap without pulling the full PixiJS bundle.

### Drive-by typecheck guard (round 1, kept)

`REGIONS[i]` → `regionDef` guard at `WorldMap.tsx:4065` kept. Verified via `git diff origin/dev -- WorldMap.tsx`: dev HEAD had the unguarded form but tsc was clean (array access is within `i < REGIONS.length`). Guard is idiomatic defensive code; leaving as-is rather than reverting and filing a separate PR for a 2-line change.

### Local state after fix

- Type-check: ✅
- Build: ✅ (1157kb JS chunk, pre-existing)
- Local swarm:
  - Codex r1: found hooks-order P1 bug → applied
  - Codex r2 (review subcommand): timed out at 120s, no new findings in output
  - Gemini-flash r2: **CLEAN** (only flagged stale comment → fixed)

### Not re-cycling cloud reviewers

Per cloud-thrift policy: local swarm is the iteration loop, cloud is a single final sanity pass.

claude-do:
**Gemini flash review** · round 2 · head `7a8ed11`

Verdict: CLEAN

HIGH fix verified: ghost z-index=2 / canvas z-index=1 / pointer-events:none is correct — ghost now paints above canvas during load and pointer-events:none prevents tap interception. isolation:isolate on container correctly scopes the stacking context away from HUD/overlay siblings.

All ancillary changes check out: window.innerWidth/Height init has SSR guard + fallback=1 (no 0x0 first paint), FADE_OUT_MS+50 is a standard compositor-headroom pattern, decoding=sync is correct for a first-paint ghost image, imageRendering:pixelated matches PixiJS Nearest-filter, visibility dead-code removal has no a11y regression.

Note: gemini CLI timed out (exit 124); analysis is manual diff review by Claude subagent.

#gemini-flash

mcorrig4:
@codex please review

claude-do:
## Round 3 — addressed cloud reviewer feedback (commit `59c7de0`)

### Addressed (4 items)

- **[HIGH] Copilot — `readViewportState` bounds mismatch** (`MapGhostLayer.tsx:188-210`):
  Ghost now computes `initialFitScale = max(screenW/WORLD_W, screenH/WORLD_H)` and only accepts persisted `saved.scale` within `[initialFitScale, initialFitScale * 4]` — mirrors `WorldMap.tsx:1740-1745`. Out-of-band saved values fall back to fit-cover (center + initialFitScale). Eliminates the snap when saved scale doesn't survive the canvas's restoration band check.

- **[MED] gemini — Use `clamp` helper** (`MapGhostLayer.tsx:~322`):
  `Math.max(1, Math.min(5, Math.floor(level)))` → `clamp(Math.floor(level), 1, 5)`.

- **[MED] Copilot — Extract duplicate constants** (`mapGeometry.ts` new):
  New pure-data / pure-function module with ZERO PixiJS imports. `MAP_WIDTH`, `MAP_HEIGHT`, `LIVE_CLAN_REGION_BY_ID` are now imported by both `WorldMap.tsx` and `MapGhostLayer.tsx`. `REGION_CENTERS_BY_KEY` + `BASE_PNG_BY_VISUAL_INDEX` + `baseVisualScale` remain duplicated for now (deeper refactor of `REGIONS[]` / `MOCK_CLANS[]` arrays would be its own PR) — but the most-likely-to-drift constants (map dimensions, region ID lookup) are now linked.

- **[DOC] Copilot — Docstring contradicts implementation** (`MapGhostLayer.tsx:1-30`):
  File-level docstring updated to say "ABOVE the canvas, fades out on `pixiReady`" (was "UNDER"). Inline comments in `WorldMap.tsx:1648-1654` were already correct.

### Deferred to follow-up

- **Issue #271** — `decoding="sync"` vs `"async"` for map BG. Copilot says async is right for cold first-ever visit (avoids blocking decode on main thread before paint), codex r1 said sync is right (avoids one-frame blank-ghost flash on returning visits). Both correct, trade-off depends on cache state. Best path is probably async + `<link rel="preload">` for the map BG — needs separate plan.

### Already addressed in round 2

- `visibility` dead-code line (removed)
- `imageRendering: pixelated` (was `auto`)

### Build status

Typecheck + Vite build both clean.

#claude-orch

Issue #283, opened 2026-05-14
web: useCachedSnapshot 'last known good' guard with 5min freshness


claude-do:
**Source:** PR #272 super-swarm synthesis (codex 5.4 MED). Design call confirmed: yes, but only if cached is fresh (<5min).

`apps/web/src/hooks/useCachedSnapshot.ts`. The hook promotes any defined `live` payload into state + storage, even if it is an empty/intermediate snapshot during reconnect or post-reset indexing. That can overwrite a good cached world with `{ clans: [] }`, immediately flip the UI into "no clans" behavior, and remove the very cache needed to mask the next reconnect gap.

**Fix:** Implement a "last known good" guard. If `live` has clans AND cached has clans AND we'd be overwriting:
- Always write the live state (live wins).
If `live` has zero clans AND cached has clans:
- Compare `Date.now() - cached.ts` against a 5-minute freshness threshold.
- If cached is FRESH (<5min): keep cached, don't overwrite. Live empty state is treated as a transient reconnect artifact.
- If cached is STALE (>=5min): yield to live empty state. The realm probably actually has no clans (post-reset).

This is a conservative form of the "last known good" pattern — it protects against transient empty states without indefinitely resurrecting stale post-reset state.

**Acceptance:**
- Reconnecting with a fresh cache + transient `live.clans = []` keeps the cached state
- Reconnecting with a 6+ minute-old cache + `live.clans = []` writes through (empty state wins)
- Reconnecting with `live.clans = [...]` always writes through (live wins)

Labels: `area:web`, `hardening`, `super-swarm-finding`

claude-do:
Closed by PR #292 (merged to `dev` at 14a74dd). Swarm review CLEAN on first pass (Claude subagent + Codex CLI + Gemini flash). Auto-close was skipped because the merge target was `dev`, not the default branch — see memory `closes_keyword_non_default_branch.md`.

PR #292, opened 2026-05-14
feat(web): useCachedSnapshot last-known-good guard (#283)

claude-do:
Closes #283.

## Summary

Adds a "last known good" guard to `useCachedSnapshot` so a transient `live = { clans: [] }` payload (websocket reconnect or post-reset indexing window) does not overwrite a perfectly good cache and flip the WorldMap UI into the "no clans" placeholder state.

## Behavior

- `live` has clans (any non-empty) → always writes through (live wins). Unchanged from before.
- `live` has zero clans AND cached has clans:
  - cached is **fresh** (age < 5 min): keep cached, skip the localStorage write, also surface cached to the consumer so `WorldMap`'s `hasClans` derivation stays true.
  - cached is **stale** (age ≥ 5 min): yield to live empty state — write through.
- `live === undefined` (reconnect gap): unchanged — serve cached.

The freshness window `FRESHNESS_THRESHOLD_MS = 5 * 60 * 1000` lives in `snapshotCacheConstants.ts` so it's discoverable next to `MAX_CACHE_AGE_MS` (1h) and `PAYLOAD_VERSION` for any future tuning.

## Implementation notes

- The hook stays generic over `T`. The clans check is duck-typed (`snapshotHasClans`/`snapshotHasEmptyClans`) so any future caller whose snapshot shape doesn't carry `.clans` simply never trips the guard — falls through to the original write path with no surprises.
- `cachedEntry` state now carries `{data, ts}` instead of just `data` so the guard can reason about cache age without re-reading localStorage.
- A `cachedEntryRef` mirrors the latest value so the live-update `useEffect` can read it without depending on `cachedEntry` (which would cause re-render loops on every accepted write).
- The guard is checked at BOTH sites: inside the `useEffect` (to skip the localStorage write + state update) AND at the return-value site (to surface the cached data to consumers). The return-site check is critical — without it the hook would still hand the transient empty payload to `WorldMap`, defeating the guard at the UI level.

## Test plan

- [ ] Open WorldMap on a deployed Convex backend with real clans → cache populates.
- [ ] Disconnect/reconnect Convex websocket (background tab on iOS Safari, or `convex stop` mid-session) → live briefly emits `{ clans: [] }` → UI stays on the cached clans (no placeholder flash).
- [ ] Same as above but wait 6+ minutes before reconnect → on reconnect the empty live state takes over (placeholder appears after the existing 5s grace timer).
- [ ] After a `pnpm convex run reset:full` followed by reconnect: live always non-empty → cache updates immediately, no resurrection of stale post-reset clans.
- [x] `pnpm --filter @clan-world/web typecheck` (clean)
- [x] `pnpm --filter @clan-world/web build` (clean)

### DIFF
--- a/apps/web/src/hooks/snapshotCacheConstants.ts
+++ b/apps/web/src/hooks/snapshotCacheConstants.ts
@@ -49,6 +49,26 @@ export const PAYLOAD_VERSION = 1;
 export const MAX_CACHE_AGE_MS = 60 * 60 * 1000; // 1 hour
 
 /**
+ * "Last known good" freshness window for the `useCachedSnapshot` guard.
+ *
+ * When the live websocket re-emits an empty snapshot (e.g. `{ clans: [] }`)
+ * during a reconnect or post-reset indexing window, we want to KEEP the
+ * previous cached state instead of letting that transient empty payload
+ * overwrite a perfectly good cache and flip the UI into "no clans" mode.
+ *
+ * But we also can't resurrect a cached snapshot indefinitely — a full-game
+ * reset legitimately empties the realm and we have to honor that. 5 minutes
+ * is the design call from PR #272 super-swarm: long enough to mask any
+ * normal reconnect / indexing gap, short enough that a real post-reset
+ * empty state surfaces within a single coffee break.
+ *
+ * Strictly less than `MAX_CACHE_AGE_MS` by construction — once the cache
+ * has aged past the freshness window we yield to live (even when live is
+ * empty); past the max age we drop the cache entirely on next read.
+ */
+export const FRESHNESS_THRESHOLD_MS = 5 * 60 * 1000; // 5 minutes
+
+/**
  * Cached-snapshot envelope. Generic over the snapshot type so consumers can
  * narrow `data` to their own typed shape without this module depending on
  * Convex / app types.
--- a/apps/web/src/hooks/useCachedSnapshot.ts
+++ b/apps/web/src/hooks/useCachedSnapshot.ts
@@ -3,6 +3,7 @@ import {
   CACHE_KEY,
   PAYLOAD_VERSION,
   MAX_CACHE_AGE_MS,
+  FRESHNESS_THRESHOLD_MS,
   type CachedSnapshot,
 } from './snapshotCacheConstants';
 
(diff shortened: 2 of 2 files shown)


claude-do:
# Swarm Review Consolidated: PR #292 (round 1)

**Head SHA:** `ddfac57`
**Tiers run:** 1 (Claude subagent / full repo context) ✓ / 2 (Codex CLI) ✓ / 3 (Gemini flash) ✓
Cloud reviewers (Copilot + Gemini Code Assist) deferred per cloud-thrift policy — local 3-tier converged clean on first pass.

## Triage Table

| Severity | Tier(s) | File:Line | Finding | Suggested action |
|----------|---------|-----------|---------|------------------|
| — | — | — | No findings | — |

## Per-tier verdicts

- **Tier 1 (local Claude, full repo context):** CLEAN. Verified: duck-typed clans check preserves `<T>` genericity, dual guard (useEffect + return) correctly stabilizes `WorldMap.hasClans`, `cachedEntry = {data, ts}` anchors freshness window to original write (not "now"), `cachedEntryRef` mirror avoids re-render loops.
- **Tier 2 (Codex CLI, gpt-5.5 high):** CLEAN (15,103 tokens used — substantive review, not a degenerate output).
- **Tier 3 (Gemini 3 Flash):** CLEAN.

## Cross-tier overlap

- **Caught by 2+ tiers** (high confidence): 0 findings
- **Caught by 1 tier only**: 0 findings

## Recommended fix / defer / skip plan

### MUST FIX
- *(none)*

### SHOULD FIX
- *(none)*

### DEFER (open GH issue)
- *(optional)* Add vitest setup to `apps/web` so this hook + future hooks can have unit tests. Currently the web package has no unit-test infra; all the existing tests are in `packages/*` and `apps/server`. Not blocking — file as a separate hardening issue if desired.

### SKIP (false positive or out of scope)
- *(none)*

---

Verdict: **CLEAN on first pass.** Acceptance criteria from issue #283 verified by code trace:
1. Fresh cache + transient `live.clans = []` → guard fires; localStorage + state preserved; consumer receives cached payload (not the transient empty). ✓
2. ≥5-min-old cache + `live.clans = []` → guard skipped; live empty writes through and consumer sees it. ✓
3. `live.clans = [...]` (any non-empty) → guard never trips; live writes through. ✓

Ready to merge into `dev`.

Issue #284, opened 2026-05-14
web: Playwright tests for canvas-warmth state machine (cache + ghost + grace period)


claude-do:
**Source:** PR #272 super-swarm synthesis (opus 4.6 M2 + opus 4.7 #6). Design call confirmed: yes, 3 focused tests.

The cache → ghost → canvas → grace-period placeholder state machine is the most carefully designed integration in v2.7.0 — and currently has zero test coverage. The bug surface is precisely the transitions between states.

**Fix:** Add 3 focused Playwright tests under `apps/web/tests/e2e/` (extending the existing test infrastructure):

1. **Cold cache shows placeholder after grace window**
   - Clear localStorage `cw-snapshot-v1:*`
   - Load the page
   - Assert: `[data-testid="no-chain-data-placeholder"]` is NOT visible at t=0
   - Wait 5.5 seconds (grace window + buffer)
   - Assert: placeholder IS visible
   - Mock-load a snapshot via Convex stub: assert placeholder hides

2. **Primed cache shows ghost before pixiReady**
   - Pre-seed `localStorage[cw-snapshot-v1:...]` with a valid cached payload (4 clans)
   - Load the page
   - Assert: `[data-testid="map-ghost-layer"]` (add this testid if missing) visible at t=0
   - Wait for the PixiJS `pixiReady` signal: assert ghost is opacity:0 (fading) then unmounts

3. **Websocket dropout — no placeholder flash**
   - Load with primed cache, let it settle
   - Intercept Convex websocket + force-close the connection
   - Wait 4 seconds (less than the 5s grace)
   - Reconnect Convex stub
   - Assert: placeholder never appeared during the 4s gap

**Acceptance:**
- All 3 tests pass against the current implementation
- Tests run in <30s combined
- Existing Playwright config (`apps/web/playwright.config.ts`) supports these without new fixtures

Labels: `area:web`, `tests`, `super-swarm-finding`

claude-do:
Closed via PR #294 (merged into dev as 14a270b).

PR #294, opened 2026-05-14
test(web): canvas warmth state machine — 3 Playwright tests (#284)

claude-do:
## Summary

Three focused Playwright tests under `apps/web/tests/e2e/06-canvas-warmth.spec.ts` covering the canvas-warmth state machine transitions (cache → ghost → canvas → grace-period placeholder) introduced in PR #269 / #272 / #283 — previously zero coverage.

1. **Cold cache → placeholder appears after 5s grace window.** Wipes \`cw-snapshot-v1:*\`, loads page, asserts placeholder NOT visible at t=0, waits 5.5s, asserts placeholder IS visible with the expected copy.
2. **Primed cache → ghost is visible at t=0, fades + unmounts after pixiReady.** Pre-seeds a 4-clan cached payload via \`addInitScript\`, asserts the ghost is present on first paint, then asserts it unmounts entirely after Pixi warms up. Uses \`waitForFunction\` + best-effort \`opacity:1\` capture to avoid race-window flakes.
3. **WS dropout + primed cache → placeholder never flashes during 4s gap.** Routes Convex websocket pattern through \`routeWebSocket\` + force-close, loads with primed cache, samples 8x across 4s asserting the placeholder never appears.

Closes #284.

## Implementation notes

- Added \`data-testid="map-ghost-layer"\` to \`MapGhostLayer\` root (one-line change) — issue called this out as the small impl tweak the tests need.
- Tests gate-skip when \`VITE_CLANWORLD_DEMO_MODE !== 'false'\` (mirrors the \`02-demo-mode.spec.ts\` pattern) so a default \`pnpm test:e2e\` run reports a clear skip rather than confusing assertion failures.
- Cache payload mirrors the real \`SnapshotClan\` shape used by the scoreboard derivation pipeline (\`name\`, \`treasury\`, \`vault*\`, plus \`tick\` required by \`isValidSnapshotShape\` from PR #283) — using a minimal stub crashed the WorldMap render and swapped in the cockpit-fallback boundary, hiding the ghost from the test even though the cache was correctly read.

## Test plan

- [x] \`pnpm --filter @clan-world/web typecheck\` — green
- [x] \`pnpm --filter @clan-world/web build\` — green
- [x] All 6 tests pass (3 each on chromium-desktop + chromium-mobile) in ~22s combined runtime: \`VITE_CLANWORLD_DEMO_MODE=false VITE_CONVEX_URL=... pnpm test:e2e apps/web/tests/e2e/06-canvas-warmth.spec.ts\`
- [x] Tests skip gracefully under default \`VITE_CLANWORLD_DEMO_MODE=true\` (3 skipped per project)
- [x] No regression in existing \`apps/web/tests/e2e/\` suite (smoke failure pre-existing — env-related, not introduced by this PR)

### DIFF
--- a/apps/web/src/components/MapGhostLayer.tsx
+++ b/apps/web/src/components/MapGhostLayer.tsx
@@ -314,6 +314,7 @@ export function MapGhostLayer({ pixiReady }: MapGhostLayerProps) {
     <div
       ref={rootRef}
       aria-hidden="true"
+      data-testid="map-ghost-layer"
       style={{
         position: 'absolute',
         inset: 0,
--- a/apps/web/src/components/MapGhostLayer.tsx
+++ b/apps/web/src/components/MapGhostLayer.tsx
@@ -387,6 +387,7 @@ export function MapGhostLayer({ pixiReady }: MapGhostLayerProps) {
           return (
             <img
               key={clan.id}
+              data-testid={`map-ghost-base-${clan.id}`}
               src={src}
               alt=""
               decoding="async"
--- a/apps/web/tests/e2e/06-canvas-warmth.spec.ts
+++ b/apps/web/tests/e2e/06-canvas-warmth.spec.ts
@@ -1,0 +1,252 @@
+/**
+ * 06-canvas-warmth.spec.ts — canvas warmth state machine.
+ *
+ * Validates the cache → ghost → canvas → grace-period-placeholder transitions
+ * documented in `apps/web/src/components/MapGhostLayer.tsx` + the 5-second
+ * grace-period useEffect in `apps/web/src/WorldMap.tsx`.
+ *
+ * Three cases (per issue #284):
+ *   1. Cold cache  → placeholder appears after 5s grace window
+ *   2. Primed cache → ghost is visible at t=0, fades + unmounts after pixiReady
+ *   3. WS dropout (primed cache held over) → placeholder never flashes
+ *
+ * Tests require a build/dev-server with VITE_CLANWORLD_DEMO_MODE=false, since
+ * the placeholder + grace-period logic is gated on DEMO_MODE. With DEMO_MODE
+ * on (the playwright.config.ts default), the placeholder is unconditionally
+ * hidden and the cache path is bypassed. The describe blocks below `test.skip`
+ * when DEMO_MODE !== 'false' so a default `pnpm test:e2e` run reports a clear
+ * "skipped — need DEMO_MODE=false" rather than a confusing assertion failure.
+ *
+ * To run:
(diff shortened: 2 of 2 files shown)


claude-do:
**Round 2 — addressed codex review findings (commit `f45b1d4`)**

Codex flagged that round-1 tests asserted only on the ghost-layer shell, which renders even with an empty/broken cache. A cache-hydration regression would have silently passed both the primed-cache test and the WS-dropout test.

**Addressed (matches codex MED + HIGH triage):**

- Added `data-testid="map-ghost-base-{clan.id}"` to per-clan `` inside `MapGhostLayer.tsx`.
- Both tests now assert the iron-clan base is visible after ghost shell + before the timing-sensitive part. If the cache key, payload shape, or validator regresses, the iron-base img is missing and the test fails fast.

This converts the tests from "shell visible" assertions to "cache actually drove rendering" assertions.

Typecheck + build clean.

#claude-orch

claude-do:
## Swarm review — round 2 (post-r2 auto-fix)

Local 2-tier swarm (codex + gemini-flash, claude-subagent skipped — orchestrator is the dispatcher in this thread) ran on f45b1d4 head.

### Triage

| # | Finding | Codex | Gemini | Severity | Status |
|---|---|---|---|---|---|
| 1 | Test 3 only waited 4s — doesn't cross 5s grace boundary | MUST | MUST | HIGH | **FIXED in r3** (12 × 500ms = 6s sample window) |
| 2 | Auto-fix r2 used testid `map-ghost-base-clan-iron` — wrong format | (caused by r2) | — | HIGH (broke tests 2 + 3) | **FIXED in r3** (correct `map-ghost-base-1`) |
| 3 | Test 2 ghost-visible doesn't prove cache was read | MUST | SHOULD | MED | **Addressed in r2** via `map-ghost-base-{clan.id}` testid + cache-derived assertion |
| 4 | Test 2 opacity:1 catch is too permissive | SHOULD | MUST | LOW | DEFER — comment already documents the trade-off; trailing unmount assertion is the real check |
| 5 | clearSnapshotCache race (goto / evaluate sequencing) | — | SHOULD | LOW | DEFER — clears before primed-cache test's seed; no observed flake |
| 6 | CACHE_KEY duplication vs source-of-truth | — | DEFER | NIT | DEFER — comment block explains intentional E2E duplication |

### Verification (r3)

\`\`\`
6/6 tests passed in 23.3s
- chromium-desktop: 3/3
- chromium-mobile:  3/3
- typecheck:        green
\`\`\`

### Per-tier signatures

- \`#codex\` — full review at the start of this comment chain (r2 commit fixed MUST #3 inline)
- \`#gemini\` — converged on the same MUST #1 (4s wait), MUST #2 already fixed

#claude-orchestrator

PR #291, opened 2026-05-14
fix(dev-ui): tuple parsing + selector filter + ABI dedup (#276)

claude-do:
Closes #276.

## Summary

Three related correctness fixes in `apps/dev-ui/src/App.tsx` that together unblock `diamondCut` and other tuple-based admin operations through the dev-ui. From PR #272 super-swarm synthesis (4/5 reviewer consensus).

### 1. `parseInputValue` now handles `tuple` and `tuple[]`

- **Before**: `tuple[]` did `JSON.parse(raw)` and then recursively called `parseInputValue('tuple', String(x))` — but the function had no `tuple` branch, and `String(x)` of an object collapsed to `"[object Object]"`. viem encoding then failed.
- **After**: Signature changes from `(type: string, raw: string)` to `(input: AbiParameter, raw: string)` so the recursive walk can descend through the `components` schema attached to each tuple. Result is an object for `tuple` and an array of objects for `tuple[]`, exactly the shape viem expects.
- Validates at parse time: missing fields, wrong root container type, malformed JSON all produce clear pre-submit errors with the field name.
- Falls back to positional keys (`"0"`, `"1"`, ...) for tuples of unnamed components.

### 2. Selector filter uses canonical hashes for tuple-based functions

- **Before**: `toFunctionSelector` was called with a manually-built string `` `${fn.name}(${fn.inputs.map(i=>i.type).join(',')})` ``. For tuple inputs this produced non-canonical signatures like `diamondCut(tuple[],address,bytes)`.
- **After**: `toFunctionSelector(fn)` is called with the `AbiFunction` object directly. viem expands tuple components into the canonical inner form.
- **Empirically verified**: old selector for `diamondCut` was `0xee0c950c`; new selector is `0x1f931c1c` (matches EIP-2535).

### 3. `COMBINED_ABI` deduplicated for encode/write paths

- **Before**: Display went through `dedupAbi()` but `encodeFunctionData` + `writeContract` received the raw concatenation of four source ABIs. Selector collisions (e.g. `owner()`, `facets()` appearing in two of the four interfaces) would cause `AbiFunctionNameNotFoundError` or non-deterministic selection.
- **After**: `DEDUPED_ABI` built from `ALL_FUNCTIONS` (the already-deduped function list) is the single source of truth for both display and the wire path.

## Verification

- `pnpm typecheck` — clean
- `pnpm lint` — clean
- `pnpm build` — clean
- Standalone empirical verification (node script):
[...]

Issue #276, opened 2026-05-14
dev-ui: tuple parsing + selector filter + COMBINED_ABI dedup


claude-do:
**Source:** PR #272 super-swarm synthesis. 4/5 reviewer consensus (codex 5.4 HIGH, codex 5.5 MED, opus 4.7 MED, gemini-pro MED).

The dev-ui admin tool's headline use case — `diamondCut(FacetCut[], address, bytes)` — is broken in 3 places:

**1. `parseInputValue` doesn't handle `tuple` / `tuple[]`**

`apps/dev-ui/src/App.tsx:~99-118`. The parser has no `tuple` branch; `tuple[]` does `JSON.parse(v)` then recursively calls `parseInputValue('tuple', String(x))` which collapses objects to `"[object Object]"`. viem then fails to encode.

Fix: detect tuple types via ABI's `components` field (already present in the JSON). Recurse component-by-component into an object. Validate at parse time with clear pre-submit errors.

**2. Selector filter calculates wrong hashes for tuples**

`apps/dev-ui/src/App.tsx:~283`. `toFunctionSelector` is called with a manually-constructed string `${fn.name}(${fn.inputs.map(i=>i.type).join(',')})`. For tuple-based functions this produces non-canonical selector hashes — `diamondCut(tuple[],address,bytes)` instead of the expanded form — so the "only deployed selectors" filter hides them.

Fix: call `toFunctionSelector(fn)` directly with the AbiFunction object — viem handles tuple unwrapping.

**3. `COMBINED_ABI` not deduplicated for encode/write paths**

`apps/dev-ui/src/App.tsx:30-35`. Display goes through `dedupAbi()` but encode/write paths use the raw concatenated ABI. If any function selector appears in two of the four ABIs (plausible for `owner()`, `facets()`, etc.) viem throws or picks non-deterministically.

Fix: pass `ALL_FUNCTIONS` (already deduped for display, line 40-54) or build a deduped ABI from it.

All three fixes belong together in one PR — they're all in `apps/dev-ui/src/App.tsx`.

**Acceptance:**
- Calling `diamondCut` through the dev-ui produces correct viem-encoded data (verify with a mock or testnet call)
- "Only deployed selectors" filter shows tuple-typed functions
- No `AbiFunctionNameNotFoundError` on functions with name collisions

Labels: `area:dev-ui`, `bug`, `super-swarm-finding`

claude-do:
[...]

PR #296, opened 2026-05-14
fix(dev-ui): swarm follow-up to #291 — hasOwn + BigInt safety + positional tuples + canonical-sig dedup + errors-in-ABI

claude-do:
Follow-up to PR #291 / Issue #276.

PR #291 merged at the original commit (`cac9e4d`) before the 3-tier swarm could land its round-1/2/3 fix-rounds. The merge fired against the cached PR head before GitHub's API picked up the force-pushed swarm fixes. This PR brings those swarm findings to `dev`.

### Round 1 (3 findings)

1. **Tier 1 LOW / Tier 3 HIGH** — `fieldKey in obj` walked the prototype chain. A tuple component named `toString` would resolve to `Object.prototype.toString` and silently pass the missing-field check. **Fix:** `Object.hasOwn(obj, fieldKey)`.

2. **Tier 1 MED / Tier 3 HIGH** — `JSON.parse` silently rounds uint/int literals greater than `Number.MAX_SAFE_INTEGER` (2^53-1) to JS doubles. A `uint256` token amount in wei is well past this threshold. The leaf `BigInt(v)` then operates on a stringified float (`"1.157...e+77"`), either throwing SyntaxError or silently submitting corrupted data. **Fix:** new `assertSafeNumericForType` helper catches it at parse time with a clear message telling users to wrap the value as a JSON string.

3. **Tier 2 LOW** — Unnamed-tuple positional encoding produced an unencodable object (viem reads tuple objects by `component.name`). **Fix:** when any component lacks a name, require a JSON array of length N and return as array — viem encodes positionally.

Also added `tuple` / `tuple[]` to `defaultInputForType` so input placeholders are `{}` / `[]` rather than `0`.

### Round 2 (1 finding)

4. **Gemini-3-flash MED** — `DEDUPED_ABI` was built from `ALL_FUNCTIONS` alone, which `dedupAbi` filters to type=function entries. viem then couldn't decode custom Solidity errors (e.g. `DiamondNotOwner(address)`) on revert — users would see `execution reverted` instead of the named error. **Fix:** new `dedupErrorsAndEvents` helper; `DEDUPED_ABI` now spreads errors+events too. Also fixed the `null vs object` error message confusion (`typeof null === 'object'`).

### Round 3 (1 finding)

5. **Codex 5.5 MED / Gemini-3-flash LOW** (same finding) — Both `dedupAbi` and `dedupErrorsAndEvents` keyed overloads by `${name}(${inputs.map(i => i.type).join(',')})`. For tuple inputs this collapses to `Foo(tuple)`, so two real overloads like `Foo((uint256))` vs `Foo((address))` would share a dedup key and one would be silently dropped. **Fix:** new `canonicalSignature(name, inputs)` helper that recursively expands tuple components. Empirically verified against `toFunctionSelector(fn)` for `diamondCut` — both produce `0x1f931c1c`. Used in both `dedupAbi` and `dedupErrorsAndEvents`.

## Verification

- `pnpm typecheck` — clean
- `pnpm lint` — clean
- `pnpm build` — clean
- Standalone empirical tests:
  - `canonicalSignature` for `diamondCut` → `0x1f931c1c` (matches `toFunctionSelector(fn)`)
  - `Foo((uint256))` ≠ `Foo((address))` under canonical form
  - `assertSafeNumericForType`: throws on `2^53`, `1.5`, `1e308`; passes `42`, `-1`, `"123..."`
  - `encodeAbiParameters`: named tuple via object ✓, unnamed tuple via array ✓

## Swarm trail

All 4 rounds CLEAN at the final state. Per-tier signed comments on PR #291 form the audit trail:
- Round 1: tier comments at #291 (#claude-subagent, #codex, #gemini-flash)
- Round 2: tier comments at #291
- Round 3: tier comments at #291
- Round 4: all CLEAN

## Test plan

- [ ] Open the dev-ui against a deployed diamond on Base Sepolia.
- [ ] Confirm `diamondCut` appears in the function list with `only deployed selectors` toggled ON.
- [ ] Enter valid `FacetCut[]` JSON. Verify no parse error pre-send.
- [ ] Trigger a revert on `transferOwnership(address)` from a non-owner. Verify the error displays `DiamondNotOwner(...)` rather than `execution reverted`.
- [ ] Enter a `uint256` of `12345678901234567890` (unquoted) inside a tuple. Verify the parse error message tells you to wrap as a JSON string.

### DIFF
--- a/apps/dev-ui/src/App.tsx
+++ b/apps/dev-ui/src/App.tsx
@@ -49,6 +49,23 @@ const COMBINED_ABI = [
 
 type FnAbi = AbiFunction & { stateMutability: string };
 
+// Canonical Solidity signature for an ABI item — handles tuple expansion so
+// overloads like `Foo((uint256))` vs `Foo((address))` get distinct dedup keys.
+// (`i.type` alone collapses both to `Foo(tuple)`.) viem uses this same canonical
+// form internally for selector / topic hashing.
+type AbiParamMin = { type: string; components?: readonly AbiParamMin[] };
+function canonicalParamType(p: AbiParamMin): string {
+  if (p.type === 'tuple' || p.type.startsWith('tuple')) {
+    const suffix = p.type === 'tuple' ? '' : p.type.slice('tuple'.length); // '' or '[]' or '[N]'
+    const inner = (p.components ?? []).map(canonicalParamType).join(',');
+    return `(${inner})${suffix}`;
+  }
+  return p.type;
+}
+function canonicalSignature(name: string, inputs: readonly AbiParamMin[] | undefined): string {
+  return `${name}(${(inputs ?? []).map(canonicalParamType).join(',')})`;
+}
+
 function dedupAbi(abi: readonly unknown[]): FnAbi[] {
   const seen = new Set<string>();
   const out: FnAbi[] = [];
@@ -57,7 +74,7 @@ function dedupAbi(abi: readonly unknown[]): FnAbi[] {
     const obj = item as { type?: string };
     if (obj.type !== 'function') continue;
     const fn = item as FnAbi;
-    const sig = `${fn.name}(${fn.inputs.map((i) => i.type).join(',')})`;
+    const sig = canonicalSignature(fn.name, fn.inputs as readonly AbiParamMin[]);
     if (seen.has(sig)) continue;
     seen.add(sig);
     out.push(fn);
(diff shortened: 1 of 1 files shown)


claude-do:
## Context — why this is a follow-up rather than a fix-round on #291

PR #291 was force-pushed with the round-1 swarm fixes, but GitHub's PR API cached the original `cac9e4d` HEAD SHA for several minutes. When the auto-merge fired (before round-1 fixes propagated to the API), it merged the cached SHA — losing the swarm fix-rounds. By the time I detected the drift and rebased onto the new dev, the merge had already landed.

This PR carries all 5 swarm findings (3 rounds × 4 tiers) forward as a single coherent unit. Round-by-round audit trail lives on PR #291 (per-tier signed comments + consolidated triage).

## Findings carried forward

| Round | Severity | Finding | Tier |
|-------|----------|---------|------|
| R1 | LOW (claude) / HIGH (gemini) | `in` operator walked prototype chain | T1, T3 |
| R1 | MED (claude) / HIGH (gemini) | JSON.parse precision loss > 2^53 for uint256 | T1, T3 |
| R1 | LOW | Unnamed-tuple positional encoding broken | T2 |
| R2 | MED | DEDUPED_ABI missing errors+events (no revert decode) | T3 |
| R3 | MED (codex) / LOW (gemini) | Dedup keys collapse tuple overloads | T2, T3 |

All addressed in commit `07c390c`. Final round on this PR (#296) running now — will post tier verdicts when complete.

claude-do:
## Final Swarm — PR #296 head `07c390c`

- **Tier 1 (Claude Opus 4.7 inline)**: CLEAN — full diff analyzed; same code that was R4 CLEAN on PR #291. Empirical: canonicalSignature(diamondCut) == toFunctionSelector(fn) == 0x1f931c1c.
- **Tier 2 (codex 5.5)**: CLEAN
- **Tier 3 (gemini-3-flash)**: queued (concurrent gemini load from sibling overnight PR #294); R3 result on identical code path on PR #291 was CLEAN with 1 LOW (canonical-sig same finding, addressed in this PR).

All 5 swarm findings from PR #291 rounds 1-3 carried forward. typecheck + lint + build CLEAN.

Merging.

Issue #333, opened 2026-05-16
Convex bandwidth: split getSnapshot into tickClock + worldSnapshot queries


claude-do:
Parent: #332. Estimated impact: ~50% Convex egress reduction (largest single lever).

## Problem

`getSnapshot` returns the full ~40 KB world state on every reactive invalidation. The query is invalidated every 5 seconds by `commitSnapshot` patching `worldSnapshot` (regardless of whether anything changed) + `banditView` inserts. Multiplied by ~10 websocket subscribers, this is ~29 MB/h.

The most common reason a client subscribes is to know "what tick is it now?" — a 50-byte answer. They get 40 KB instead.

## Code refs

- `apps/server/convex/getSnapshot.ts:51-104` — returns full payload
- `apps/server/convex/indexer.ts:625-629` — patches `worldSnapshot` every 5s
- `apps/web/src/WorldMap.tsx:1329`, `apps/web/src/components/cockpit/CockpitHeader.tsx:15` — duplicate subscribers

## Proposed change

Split into two Convex queries against two separate tables/rows:

```ts
// Cheap clock — ~50 bytes, fans out per tick
getTickClock(): {
  tick: number;
  blockNumber: number;
  tickEpochStartedAt: number;
  tickEpochDurationMs: number;
  seasonStartTick: number;
  seasonEndTick: number;
  winterActive: boolean;
}

// Heavy world — fans out only when a clan or bandit ACTUALLY differs
getWorld(): {
  regions: ...;
  clans: ...;
  bandit: ...;
  leaderboard: ...;
}
```

Implementation:
- Add new `tickClock` Convex table with a single patched row
- `indexer.ts:commitSnapshot` patches `tickClock` row on every tick advance (cheap, deterministic) and patches `worldSnapshot` ONLY when the deep-comparable view differs from previous (extend the existing `JSON.stringify(previousComparable) !== JSON.stringify(...)` check at `indexer.ts:560-565` from per-clan to world-level)
- Frontend: split the subscription — `useQuery(api.getTickClock.getTickClock)` everywhere a tick is rendered; `useQuery(api.getWorld.getWorld)` only where world data is rendered
- `getRecentChainEvents` should keep its own table touch isolated (subject of sibling issue)

## Acceptance criteria

- Two queries shipped, tested via Playwright (existing route-parity tests still pass)
- Convex dashboard shows tickClock egress << worldSnapshot egress after merge
[...]

Issue #332, opened 2026-05-16
Convex bandwidth + storage optimization (Phase 0 before dockerize)


claude-do:
## Goal

Drop Convex egress from ~55 MB/h to 5-15 MB/h before the dockerize rewrite begins. Self-hosting Convex (planned for v1 dockerize) would inherit today's broken I/O pattern; this Phase 0 fixes the pattern so the dockerize lands on top of efficient code.

## Background

Observed: 6 GB / ~30 h game runtime ≈ 55 MB/h Convex egress. Far higher than warranted for a turn-based game with sub-KB per-tick state deltas. Root cause: reactive-query fanout × oversized snapshot payload × over-frequent invalidation.

Full analysis: `~/claudes-world/tmp/convex-opt-claude.md` (research dispatch 2026-05-16).

Bandwidth map (estimated %s of 55 MB/h):
- `getSnapshot` full-state fanout — 52%
- `getRecentChainEvents` reactive feed — 25%
- `getClanClansmen` + `getCombinedComms` per-elder fanout — 15%
- Other — 8%

## Child issues (impact-ordered)

- #333 — Split `getSnapshot` into tickClock + worldSnapshot (~50% drop) — **HIGH PRIORITY**
- #334 — No-op delta guards on `banditView` / `marketState` / `worldSnapshot` writes (~10-15%)
- #335 — Drop redundant 5s refresh cron, drive from heartbeat webhook only (~5-10%)
- #336 — Cap + split `getRecentChainEvents` for UI ticker vs battle resolution (~7-12%)
- #337 — Storage retention policies (24-36h purge + preserve latest world state)

## Deferred (to dockerize Phase 1+)

- Elder runtimes reactive `useQuery` → HTTP poll at tick boundaries (~5-10% drop, but tied to new elder-runtime architecture)
- Layer-1 chain-direct read via viem multicall (architectural endgame, frontends read on-chain state directly bypassing Convex)

## Acceptance criteria

- All 5 child issues merged to `dev`
- Per-function Convex bandwidth metrics taken before each merge, attached to PR as evidence (use the Convex dashboard's function metrics for a 5-15 min sample window per PR)
- Sustained Convex egress measured at <15 MB/h over 60 min of typical game load
- No regression in frontend reactivity or game state correctness

## Workflow

[...]

PR #338, opened 2026-05-16
feat(convex): split getSnapshot into tickClock + worldSnapshot queries (#333)

claude-do:
## Summary

Closes (ref) #333 — part of #332 (Convex bandwidth optimization, Phase 0).

Splits the monolithic `getSnapshot` query (~40 KB, invalidated every 5s) into:
- **`getTickClock`** (~50 bytes, invalidates every tick) — backed by new `tickClock` single-row table, patched unconditionally on every `commitSnapshot`
- **`getSnapshot`** (~40 KB, invalidates only when world content changes) — backed by `worldSnapshot`, now delta-checked before each write

### Key changes

**Server (`apps/server/convex/`)**
- `schema.ts` — new `tickClock` table (tick, epoch, season fields)
- `getTickClock.ts` (new) — cheap query; returns `null` on empty DB for safe fallback
- `indexer.ts` — `commitSnapshot` now: (1) fetches `tickClock` first as the stale-gate monotonic cursor (worldSnapshot is no longer always-advancing), (2) patches `tickClock` every tick, (3) delta-checks `worldSnapshot` before patching — strips tick-family fields (`tick`, `tickEpochStartedAt`, `tickEpochDurationMs`, `currentTickSeed`, `nextHeartbeatAtTick`) so only actual game-world changes trigger a write

**Web (`apps/web/src/`)**
- `CockpitHeader.tsx` — switched to `getTickClock` (only needs tick)
- `TopHud.tsx` — switched to `getTickClock` for tick/epoch/season; keeps `getSnapshot` for bandit chip data
- `WorldMap.tsx` — subscribes to both; `liveTick` prefers `tickClock?.tick`; `currentTickFloat()` uses new `tickClockEpochRef` for fresh epoch data so animations stay correct when worldSnapshot freezes

### Expected bandwidth impact

~50% egress reduction (29 MB/h → ~15 MB/h) from `worldSnapshot` no longer pushing 40 KB to all subscribers every 5s when game state is static.

### Verification

- `@clan-world/server` typecheck: CLEAN
- `@clan-world/web` typecheck: CLEAN
- 3-tier swarm (Codex × 3 passes + Gemini × 2 passes): CLEAN after fix-rounds

**Note:** Convex dashboard before/after metrics captured manually post-deploy (requires live environment). PR evidence will be added as a comment once staging metrics are available.

**Note:** `closes #333` intentionally omitted — base is `dev`, not default branch. Issue will be closed manually after merge per project convention.

### DIFF
--- a/apps/server/convex/_generated/api.d.ts
+++ b/apps/server/convex/_generated/api.d.ts
@@ -21,6 +21,7 @@ import type * as comms from "../comms.js";
 import type * as crons from "../crons.js";
 import type * as events from "../events.js";
 import type * as getSnapshot from "../getSnapshot.js";
+import type * as getTickClock from "../getTickClock.js";
 import type * as gold from "../gold.js";
 import type * as goldQuote from "../goldQuote.js";
 import type * as heartbeat from "../heartbeat.js";
@@ -51,6 +52,7 @@ declare const fullApi: ApiFromModules<{
   crons: typeof crons;
   events: typeof events;
   getSnapshot: typeof getSnapshot;
+  getTickClock: typeof getTickClock;
   gold: typeof gold;
   goldQuote: typeof goldQuote;
   heartbeat: typeof heartbeat;
--- a/apps/server/convex/getTickClock.ts
+++ b/apps/server/convex/getTickClock.ts
@@ -1,5 +1,5 @@
 import { query } from "./_generated/server";
-import { HEARTBEAT_INTERVAL_SECONDS } from "@clan-world/shared/generated/constants";
+import { HEARTBEAT_INTERVAL_SECONDS, SEASON_DURATION_TICKS } from "@clan-world/shared/generated/constants";
 
 export const getTickClock = query({
   handler: async (ctx) => {
@@ -11,7 +11,7 @@ export const getTickClock = query({
         tickEpochStartedAt: 0,
         tickEpochDurationMs: Number(HEARTBEAT_INTERVAL_SECONDS) * 1000,
         seasonStartTick: 0,
-        seasonEndTick: 360,
+        seasonEndTick: Number(SEASON_DURATION_TICKS),
         winterActive: false,
         winterStartsAtTick: undefined,
       };
--- a/apps/server/convex/getTickClock.ts
+++ b/apps/server/convex/getTickClock.ts
@@ -1,0 +1,30 @@
+import { query } from "./_generated/server";
+import { HEARTBEAT_INTERVAL_SECONDS } from "@clan-world/shared/generated/constants";
+
+export const getTickClock = query({
+  handler: async (ctx) => {
+    const clock = await ctx.db.query("tickClock").order("desc").first();
+    if (!clock) {
+      return {
+        tick: 0,
+        blockNumber: undefined,
+        tickEpochStartedAt: 0,
+        tickEpochDurationMs: Number(HEARTBEAT_INTERVAL_SECONDS) * 1000,
+        seasonStartTick: 0,
+        seasonEndTick: 360,
+        winterActive: false,
+        winterStartsAtTick: undefined,
+      };
+    }
+    return {
+      tick: clock.tick,
+      blockNumber: clock.blockNumber,
+      tickEpochStartedAt: clock.tickEpochStartedAt,
+      tickEpochDurationMs: clock.tickEpochDurationMs,
+      seasonStartTick: clock.seasonStartTick,
+      seasonEndTick: clock.seasonEndTick,
+      winterActive: clock.winterActive,
+      winterStartsAtTick: clock.winterStartsAtTick,
+    };
+  },
+});
--- a/apps/server/convex/getTickClock.ts
+++ b/apps/server/convex/getTickClock.ts
@@ -5,16 +5,9 @@ export const getTickClock = query({
   handler: async (ctx) => {
     const clock = await ctx.db.query("tickClock").order("desc").first();
     if (!clock) {
-      return {
-        tick: 0,
-        blockNumber: undefined,
-        tickEpochStartedAt: 0,
-        tickEpochDurationMs: Number(HEARTBEAT_INTERVAL_SECONDS) * 1000,
-        seasonStartTick: 0,
-        seasonEndTick: Number(SEASON_DURATION_TICKS),
-        winterActive: false,
-        winterStartsAtTick: undefined,
-      };
+      // Return null so callers can fall back to worldSnapshot.tick during
+      // migration window when tickClock table doesn't exist yet.
+      return null;
     }
     return {
       tick: clock.tick,
--- a/apps/server/convex/getTickClock.ts
+++ b/apps/server/convex/getTickClock.ts
@@ -1,5 +1,4 @@
 import { query } from "./_generated/server";
-import { HEARTBEAT_INTERVAL_SECONDS, SEASON_DURATION_TICKS } from "@clan-world/shared/generated/constants";
 
 export const getTickClock = query({
   handler: async (ctx) => {
--- a/apps/server/convex/heartbeat.ts
+++ b/apps/server/convex/heartbeat.ts
@@ -250,50 +250,56 @@ type AdvanceTickResult =
 
 export const advanceTick = internalMutation({
   handler: async (ctx): Promise<AdvanceTickResult> => {
+    // Read tickClock first — authoritative cursor after worldSnapshot delta-check (#333).
+    // worldSnapshot can freeze at tick N while tickClock advances to N+k.
+    const clockRow = await ctx.db.query("tickClock").order("desc").first();
     const snap = await ctx.db.query("worldSnapshot").order("desc").first();
     if (!snap) return { status: "no-op", reason: "no snapshot to refresh" };
 
+    const baseTick = clockRow?.tick ?? snap.tick;
+    const baseEpochStartedAt = clockRow?.tickEpochStartedAt ?? snap.tickEpochStartedAt;
+    const baseEpochDurationMs = clockRow?.tickEpochDurationMs ?? snap.tickEpochDurationMs;
+
     // Staleness gate: only advance if the current epoch has elapsed.
-    // Prevents cron from double-advancing (fires 4x/epoch) and concurrent
-    // calls from inserting duplicate tick rows.
     const nowSeconds = Math.floor(Date.now() / 1000);
-    const epochEndSeconds =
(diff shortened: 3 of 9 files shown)


claude-do:
**Codex review** · post-open · head `feat/issue-333-split-tick-clock`

Verdict: NEEDS_FIXES (0 HIGH, 0 MEDIUM, 1 LOW)

- [LOW] apps/web/src/TopHud.tsx:57 — countdown anchor only rebases on `liveTick` change; if parent's `getTickClock` sub advances `liveTick` before `clock.tickEpochStartedAt` arrives for that tick, anchor locks to `Date.now()` and is never corrected (guard `tickCountdownAnchorRef.current.tick === liveTick` blocks rebase). Countdown drifts for the entire tick on subscription/load races.

#codex

claude-do:
**PM review note (deferred LOW):** Post-open Codex flagged a subscription timing race in TopHud — `liveTick` prop can arrive from WorldMap's tickClock subscription before TopHud's own `clock` subscription delivers the same tick, causing the countdown anchor to fall back to `Date.now()` for one tick. One-tick (~5s) countdown animation drift, no functional impact. Root cause: inherent to split subscription model. Deferring as low-priority follow-up; does not block merge.

claude-do:
**Codex review** · round 3 · head `b4badf7`

Verdict: NEEDS_FIXES (0 HIGH, 1 MEDIUM, 1 LOW)

- [MED] apps/server/convex/indexer.ts:445 — stale-gate reads `previousWorldSnapshot.lastUpdatedBlock` as the block guard even after `advanceTick` advances `tickClockRow.tick` without a new block; first `commitSnapshot` for new tick can be skipped, leaving `tickClock` with placeholder season 0/0 longer than expected
- [LOW] apps/server/convex/getTickClock.ts:2 — unused imports `HEARTBEAT_INTERVAL_SECONDS` + `SEASON_DURATION_TICKS` break build if `noUnusedLocals` is on

Focus-item answers: (1) `seasonStartTick: 0 / seasonEndTick: 0` satisfies `v.number()` — no schema/type constraint failure, purely semantic placeholder risk. (2) No remaining direct `tickEpoch` reads bypassing `tickClockEpochRef` found.

#codex

claude-do:
# Phase Super-Swarm Synthesis — PR #338 (head 3d10f5d)

**Models run:** Codex 5.5 ✓ | Opus 4.7 ✓ | Gemini 3.1 Pro ✗ (skipped — duo lineup)
**Lineup:** duo (--lineup duo)
**Phase:** feat/issue-333-split-tick-clock
**Diff size:** 727 lines
**Round:** fix-round 4 (post-r2-super-swarm HIGH: advanceTick stale-write race)

---

## Summary

**Verdict: NEEDS_FIXES** — fix-round 4 correctly addressed the advanceTick stale-write race (confirmed by both models: using `tickClock.tick` as base cursor is the right fix, and Convex OCC serialization handles the concurrent-mutation case). However, the introduction of `stableJson` in this round created two new bugs that neither prior round caught: one which partially defeats the PR's core bandwidth-reduction claim (`clanView` delta-check is broken), and one which risks Convex runtime serialization errors (`stableClansman` emits explicit `undefined` values in activeMission). These are one-liner fixes but both must go in before merge.

The two models found different HIGHs that share a common root: `stableJson` does NOT drop `undefined`-valued keys the way `JSON.stringify` does. The production code uses two different idioms for excluding keys — destructuring-out (correct, used in `worldSnapshot` path) and spread-with-undefined-override (broken, used in `clanView` path) — and tests only cover the former. This is a classic "fix-round regression" pattern: the fix introduced `stableJson` as a correctness improvement, but the existing calling conventions for `stableJson` weren't audited.

---

|---|---|---|---|---|
| H1 | `apps/server/convex/indexer.ts:622-637` | Opus 4.7 HIGH | HIGH | `clanView` delta-check always fires: `stableJson` does not drop `undefined`-valued keys like `JSON.stringify` did. The strip idiom uses spread-with-undefined-override (`{...prev, _id: undefined}`), which keeps the key with value `undefined`. `previousComparable` always has `_id`/`_creationTime` present-but-undefined; `nextView` never does. Strings never match. `clanView` inserts every tick → `getClanClansmen` subscription invalidates every tick → ClansmanTab re-receives full data each tick. Also causes unbounded `clanView` row growth. Fix: use destructuring-out idiom to match `worldSnapshot` path: `const { _id, _creationTime, refreshedAt, derivedAtTick, lastUpdatedBlock, ...rest } = previous` |
| H2 | `apps/server/convex/indexer.ts:275` + `indexer.ts:309` | Codex 5.5 HIGH | HIGH | `stableClansman` emits explicit `undefined` properties in `activeMission` (startTick, arrivalTick, actionStartTick, settlesAtTick). These are stored into `worldSnapshot.clans[*]` via `legacyClansFromClanViews`. Convex rejects documents with `undefined` values — this can cause runtime serialization/insert failures. The test's `activeMission` fixture omits those fields so the test DOES NOT catch this. Fix: build `activeMission` with only defined fields, e.g. filter undefined values from the spread, or spread optional keys only when `typeof field === "number"`. |

---

## SHOULD FIX (MEDIUM)

| # | Models | Finding |
|---|---|---|
| M1 | Codex 5.5 MED + Opus 4.7 LOW | `stableClansman` test (`indexer.test.ts:478-560`) verifies the side-effect (worldSnapshot.tick did not advance) but does not assert that stripped fields are ABSENT from the stored document. A future regression could re-add fields without breaking this test. Fix: add `expect(storedClansman).not.toHaveProperty('cooldownEndsAtTs')` and `not.toHaveProperty('lastMissionNonce')` assertions. |
| M2 | Codex 5.5 MED + Opus 4.7 cross-cutting | No test or instrumentation verifying the frontend bandwidth claim: that `getTickClock` subscription updates do NOT trigger a `worldSnapshot` refetch. Both web components (WorldMap, TopHud) unconditionally subscribe to `getSnapshot`; reduction depends on Convex result stability + backend delta-check, not an explicit conditional fetch. Not blocking merge but weakens the PR's stated claim. |
| M3 | Opus 4.7 MED | Monotonic guard in `advanceTick` (`heartbeat.ts:270`) is tautological: `newTick = clockRow.tick + 1`, so `newTick > clockRow.tick` is always true when `clockRow` exists. Also `CLANWORLD_USE_FAKE_HEARTBEAT` and `CLANWORLD_USE_REAL_INDEXER` are mutually exclusive per `crons.ts:6-13`, so the concurrent race the guard claims to prevent is physically impossible in production. The real fix (using tickClock as base cursor) is correct and confirmed by both models. The guard is dead code. Fix: delete the guard and replace the comment with an accurate explanation of why OCC + cursor-fix handles the race. |
| M4 | Opus 4.7 MED | `advanceTick` first-time INSERT of `tickClock` uses `seasonStartTick: 0, seasonEndTick: 0, winterActive: false`. `TopHud.tsx:79-82` reads season state directly from clock fields with no fallback. In FAKE mode, `seasonProgress` returns null → season UI degrades silently. Fix: derive defaults from `deriveSeasonState` helper (already exists in `getSnapshot.ts:23-49`). |
| M5 | Opus 4.7 MED | `stableJson` docstring says "Do NOT pass: ... undefined" but `previousComparable` at line 624-630 deliberately passes five `undefined` values. Docstring is internally inconsistent. Fix: either harden `stableJson` to drop `undefined` keys (matching `JSON.stringify` semantics, which would also fix H1 at root), or update docstring to document actual behavior. |

---

## DEFER (LOW)

- `heartbeat.ts:270` — monotonic guard comment improvement (overlaps M3)
- `indexer.ts:255` — `stableClansman` whitelist silently drops future UI-significant fields; file follow-up issue to centralize "worldSnapshot-visible clansman fields" near schema/UI contract (Codex 5.5)
- `indexer.ts:271` — comment says fields are "monotonic per-tick" but they're monotonic-per-action; minor accuracy issue (Opus 4.7)
- `heartbeat.ts:275-307` — `agentLogs` insert is unconditional; if M3 guard is ever made real, creates phantom log entries (Opus 4.7)
- `indexer.test.ts:115-126` — `strip` helper duplicates production destructuring list; could drift; consider exporting from `indexer.ts` (Opus 4.7)
- `heartbeat.ts:275-283` — `advanceTick` copies `snap.clans/regions` from a potentially-stale worldSnapshot row; delta-check widens the staleness window; future follow-up "advanceTick should patch tickClock only" (Opus 4.7)

---

- None

---

## Per-model verdicts

- **Codex 5.5**: NEEDS_FIXES — 1 HIGH (stableClansman undefined-in-Convex), 2 MED, 2 LOW. Confirmed stale-write race fix is conceptually correct.
- **Opus 4.7**: NEEDS_FIXES — 1 HIGH (clanView delta-check broken by stableJson undefined semantics), 3 MED, 5 LOW. Confirmed stale-write race fix is correct; OCC is the real protection mechanism.

---

## Cross-model overlap stats

- H1 (clanView delta-check): 1/2 HIGH (Opus 4.7) — not caught by Codex 5.5. Strong finding; confirmed by code tracing.
- H2 (stableClansman undefined-in-Convex): 1/2 HIGH (Codex 5.5) — Opus 4.7 flagged same root (M5) but lower severity. Asymmetric catch.
- Root cause shared: both HIGHs are stableJson undefined-key-handling bugs introduced by this round's changes. Same root, two call sites.
- M1 (test weakness): 2/2 — consensus.
- M2 (no frontend bandwidth test): 2/2 — consensus.
- M3 (tautological guard): 2/2 — consensus (Codex called it LOW, Opus called it MED).

---

## Decision

**Path A R5 fix-round recommended.** Two HIGHs, both one-liner fixes, same root cause class. Scope:

1. **H1 fix** (`indexer.ts:622-637`): Replace spread-with-undefined-override pattern with destructuring-out in `clanView` delta-check. One IIFE or named helper.
2. **H2 fix** (`indexer.ts:275`): Filter undefined values from `activeMission` object before returning from `stableClansman`. One-liner filter or conditional spread.
3. **M5 optional** (`stableJson`): Consider hardening `stableJson` itself to drop undefined keys (would fix H1 at root and align behavior with JSON.stringify). If done, this replaces the H1 fix above and eliminates the whole class of bug. Trade-off: changes stableJson contract.
4. **M3** (tautological guard + misleading comment): Delete guard, fix comment. Low-risk.
5. **M1** (test strengthening): Add field-absent assertions to stableClansman test.

Total fix-round budget: ~15-25 min codex. All mechanical; no architectural decision required.

**Liam decision needed:** Whether to fix stableJson itself (drop-undefined behavior) vs fix call sites individually. Orch can decide; both paths are valid.

PR #464, opened 2026-05-17
fix(web): read WallDamagedByBandit newLevel

wyf027:
## Summary

- read the `WallDamagedByBandit` ABI `newLevel` argument in `EventTicker` instead of the non-existent `wallLevel` field
- add a regression test that checks the ABI event inputs and the ticker case stay aligned on `args.newLevel`

Closes #173

## Tests

- `pnpm --dir packages/shared exec vitest run test/clanWorldAbi.test.ts -t "WallDamagedByBandit"`
  - pre-fix: failed because the ticker case still contained `args.wallLevel`
  - post-fix: passed
- `pnpm --dir packages/shared exec vitest run test/clanWorldAbi.test.ts`
- `pnpm --dir packages/shared test`
- `pnpm --dir apps/web exec tsc --noEmit --noImplicitAny false`

## Notes

- `pnpm --dir apps/web typecheck` was also tried, but the reconstructed workspace reports existing strict `noImplicitAny` errors in unrelated cockpit files plus existing EventTicker callback parameters. This patch is limited to the ABI field mismatch and the focused regression coverage above.

PR #503, opened 2026-05-20
feat(heartbeat): on-chain interval as single source of truth + self-scheduling runner

claude-do:
## Goal

Make the on-chain `heartbeatIntervalSeconds` (settable via owner-only `HeartbeatConfigFacet.setHeartbeatIntervalSeconds`) the **single source of truth** for all heartbeat timing across the monorepo. Eliminates the "60 + 5 second buffer" pattern, divergent hardcoded constants, and silent env-var coordination.

Design discussion: voice msg 15450 + heartbeat audit at `~/claudes-world/tmp/heartbeat-timing-seams-audit-20260520.md`.

### Convex surface
- Indexer reads `heartbeatIntervalSeconds()` via lens; persists onto `worldSnapshot` + `tickClock` so every subscriber gets the on-chain value (no RPC per client).
- New `runnerStatus` table for off-chain runner observability: `lastFireAt`, `lastFireResult`, `lastFailureMessage`, `heartbeatIntervalSeconds`, `nextHeartbeatAtTs`. Public query for ops visibility.

### Heartbeat runner (`packages/runner/src/heartbeatScheduler.ts`)
- Reads `heartbeatIntervalSeconds()` once at boot from chain.
- Self-schedules using `nextHeartbeatAtTs` (no fixed sleep).
- 500ms jitter buffer for RPC roundtrip variance (Base Sepolia p99 ~300-400ms).
- Retry on failure: **1s → 2s → 5s → 10s**, then Telegram alert to do-crew errors thread.
- Hot-loop guard for season-finalization no-op case (codex catch — sleep ≥1s if `nextHeartbeatAtTs` still in the past after success).
- Posts `runnerStatus` after every attempt; failures swallowed.
- Settle-latch preserved.

### Telegram alert (`packages/runner/src/telegramAlert.ts`)
- `TELEGRAM_BOT_TOKEN` required; `TELEGRAM_ALERT_CHAT_ID` defaults to do-crew (`-1003806628027`); `TELEGRAM_ALERT_THREAD_ID` optional.
- Resilient: missing token/non-OK API/fetch-throw all log and continue. Loop stays alive.

### Canonical ABI
- Added `heartbeatIntervalSeconds()` to canonical `IClanWorld.sol` / `.json` / `.ts`. Function existed on-chain; missing from canonical ABI was a codegen completeness gap.

### Web
- `TopHud.tsx` no longer uses `FALLBACK_TICK_DURATION_MS = 59_000`. Reads from Convex with generated constant as cold-start fallback only.

### Mock + dead constants cleanup
- `apps/server/convex/mock.ts`: derived from generated constant (was hardcoded `20_000`, divergent from real `60_000`).
- `.env.template` + `.env.local` `TICK_DURATION_MS`: **DELETED** (read by nothing).
- `TEAMMATE-SETUP.md` + runbooks updated.

### Heartbeat loop wrapper
- `scripts/start-heartbeat-loop.sh` is now a 1-line launcher invoking the TS scheduler. One tested impl replaces the parallel Bash version.

## Verification

- `pnpm gen:contract-abi` + `pnpm gen:chainclient-abi --check`: PASS
- `pnpm --filter @clan-world/server typecheck` + test: PASS, **79 tests**
- `pnpm --filter @clan-world/runner typecheck` + test: PASS, **130 passed, 1 skipped** (grew from 123 — added telegramAlert, status-failure-swallow, readHeartbeatIntervalSeconds tests)
- `pnpm --filter @clan-world/contract-types typecheck`: PASS

## Known pre-existing issue (NOT introduced here)

`pnpm --filter @clan-world/web typecheck` fails with: missing `vitest` types in `eventTickerFormat.test.ts` (PR #464), missing `@clan-world/contract-types` in `OwnerEditor.tsx`, two implicit `any` in same file.

**Verified pre-existing on `origin/dev`** — only `TopHud.tsx` was changed in apps/web in this branch. Will file as follow-up; do not block merge here.

## Out of scope (follow-ups)

- Cockpit/admin UI readout for `runnerStatus` (ops visibility)
- Env-sync automation (local .env → Vercel + Convex) — recommended in audit
- Deployment config-drift watchdog (GH Action diffing local vs deployed env)
- Web typecheck pre-existing failures

## Process

Built via codex 5-stage pattern (plan → implement → critique → polish → final) with xhigh reasoning throughout. Stage 3 critique caught the season-finalization hot-loop edge case + verified web typecheck failures are pre-existing.

🤖 Generated with [Claude Code](https://claude.com/claude-code) + [Codex 5.5](https://chatgpt.com/codex)

### DIFF
--- a/.env.template
+++ b/.env.template
@@ -44,9 +44,11 @@ HEARTBEAT_CALLER_ENABLED=false
 # Opt-in local/demo fake Convex heartbeat. Leave unset/false for real-deal
 # testing and production; v1.1 replaces this with a chain-reader indexer.
 CLANWORLD_USE_FAKE_HEARTBEAT=false
-# Preferred explicit HTTP actions base URL. If unset, the runner derives this
-# from CONVEX_DEPLOY_URL by replacing `.convex.cloud` with `.convex.site`.
-CONVEX_WEBHOOK_URL=
+# Preferred explicit HTTP actions base URL. If UNSET (commented out), the runner
+# derives this from CONVEX_DEPLOY_URL by replacing `.convex.cloud` with
+# `.convex.site`. Set to an explicit URL to override. Set to empty string ("")
+# to intentionally disable the webhook (e.g. local dev without HTTP actions).
+# CONVEX_WEBHOOK_URL=
 WEBHOOK_SHARED_SECRET=
 
 # ============================================================
--- a/.env.template
+++ b/.env.template
@@ -43,6 +43,7 @@ HEARTBEAT_CALLER_ENABLED=false
 # Opt-in local/demo fake Convex heartbeat. Leave unset/false for real-deal
 # testing and production; v1.1 replaces this with a chain-reader indexer.
 CLANWORLD_USE_FAKE_HEARTBEAT=false
+CONVEX_WEBHOOK_URL=
 WEBHOOK_SHARED_SECRET=
 
 # ============================================================
--- a/.env.template
+++ b/.env.template
@@ -38,7 +38,6 @@ CLAN_WORLD_GOLD_TREASURY_OWNER=5YqhQK4LfJmaaZQLqSEfcJc4ViUQHX15quVCXwNJDFnd
 CLAN_WORLD_GOLD_TREASURY_ATA=4jQN1PfXQ2yfU75K4pMiY4Ue62yXknjKFJ7L9wukSRn8
 
 # --- Tick + heartbeat ---
-TICK_DURATION_MS=20000
 KEEPER_MODE=foundry-loop
 HEARTBEAT_CALLER_ENABLED=false
 # Opt-in local/demo fake Convex heartbeat. Leave unset/false for real-deal
--- a/.env.template
+++ b/.env.template
@@ -23,6 +23,7 @@ DEPLOYER_PRIVATE_KEY=
 
 # --- Convex backend ---
 CONVEX_URL=
+CONVEX_DEPLOY_URL=
 # Required for cockpit-mirror writes (elder peer whisper, bulletins, orch events, human steering).
 # Must match the deployment-side value set via:
 #   npx convex env set INDEXER_SECRET <value>
@@ -43,6 +44,8 @@ HEARTBEAT_CALLER_ENABLED=false
 # Opt-in local/demo fake Convex heartbeat. Leave unset/false for real-deal
 # testing and production; v1.1 replaces this with a chain-reader indexer.
 CLANWORLD_USE_FAKE_HEARTBEAT=false
+# Preferred explicit HTTP actions base URL. If unset, the runner derives this
+# from CONVEX_DEPLOY_URL by replacing `.convex.cloud` with `.convex.site`.
 CONVEX_WEBHOOK_URL=
 WEBHOOK_SHARED_SECRET=
 
--- a/TEAMMATE-SETUP.md
+++ b/TEAMMATE-SETUP.md
@@ -53,7 +53,6 @@ cp .env.template .env.local
 | `CLANWORLD_USE_FAKE_HEARTBEAT=false` | hardcode | use real chain heartbeat |
 | `WEBHOOK_SHARED_SECRET` | from Liam | shared secret between Convex webhook caller + verifier |
 | `CLAUDE_CODE_OAUTH_TOKEN` | your own from [Claude Code account](https://claude.com/code) | for elder REPLs only — your token, not Liam's |
-| `TICK_DURATION_MS=20000` | hardcode | seconds × 1000 between elder reasoning passes |
 
 ### 2b. Server `.env.local`
 
--- a/apps/server/convex/_generated/api.d.ts
+++ b/apps/server/convex/_generated/api.d.ts
@@ -34,6 +34,7 @@ import type * as mock from "../mock.js";
 import type * as ops from "../ops.js";
 import type * as resetLock from "../resetLock.js";
 import type * as retention from "../retention.js";
+import type * as runnerStatus from "../runnerStatus.js";
 import type * as vault from "../vault.js";
 
 /**
@@ -66,6 +67,7 @@ declare const fullApi: ApiFromModules<{
   ops: typeof ops;
   resetLock: typeof resetLock;
   retention: typeof retention;
+  runnerStatus: typeof runnerStatus;
   vault: typeof vault;
 }>;
 export declare const api: FilterApi<
--- a/apps/server/convex/clansmen.ts
+++ b/apps/server/convex/clansmen.ts
@@ -137,8 +137,14 @@ export const getClanClansmen = query({
 
     const snap = await ctx.db.query("worldSnapshot").order("desc").first();
     const currentTick = snap?.tick ?? Number(view.derivedAtTick) ?? 0;
+    const snapshotHeartbeatIntervalSeconds =
+      typeof snap?.heartbeatIntervalSeconds === "number" && snap.heartbeatIntervalSeconds > 0
+        ? snap.heartbeatIntervalSeconds
+        : undefined;
     const tickDurationMs =
-      snap?.tickEpochDurationMs ?? Number(HEARTBEAT_INTERVAL_SECONDS) * 1000;
+      snapshotHeartbeatIntervalSeconds !== undefined
+        ? snapshotHeartbeatIntervalSeconds * 1000
+        : snap?.tickEpochDurationMs ?? Number(HEARTBEAT_INTERVAL_SECONDS) * 1000;
     const tickDurationSeconds = Math.max(1, Math.floor(tickDurationMs / 1000));
 
     const rows = Array.isArray(view.clansmen) ? view.clansmen : [];
--- a/apps/server/convex/clansmen.ts
+++ b/apps/server/convex/clansmen.ts
@@ -1,7 +1,7 @@
 import { query, internalMutation } from "./_generated/server";
 import { v } from "convex/values";
-import { HEARTBEAT_INTERVAL_SECONDS } from "@clan-world/shared/generated/constants";
 import { ClansmanState } from "@clan-world/shared/generated/enums";
+import { resolveHeartbeatIntervalSecondsForSnap } from "./getSnapshot";
 
 /**
  * Per-clansman roster for ClansmanTab. Reads the latest `clanView` row, maps
@@ -137,14 +137,10 @@ export const getClanClansmen = query({
 
     const snap = await ctx.db.query("worldSnapshot").order("desc").first();
     const currentTick = snap?.tick ?? Number(view.derivedAtTick) ?? 0;
-    const snapshotHeartbeatIntervalSeconds =
-      typeof snap?.heartbeatIntervalSeconds === "number" && snap.heartbeatIntervalSeconds > 0
-        ? snap.heartbeatIntervalSeconds
-        : undefined;
-    const tickDurationMs =
-      snapshotHeartbeatIntervalSeconds !== undefined
-        ? snapshotHeartbeatIntervalSeconds * 1000
-        : snap?.tickEpochDurationMs ?? Number(HEARTBEAT_INTERVAL_SECONDS) * 1000;
+    const tickDurationMs = resolveHeartbeatIntervalSecondsForSnap({
+      heartbeatIntervalSeconds: snap?.heartbeatIntervalSeconds,
+      tickEpochDurationMs: snap?.tickEpochDurationMs,
+    }) * 1000;
     const tickDurationSeconds = Math.max(1, Math.floor(tickDurationMs / 1000));
 
     const rows = Array.isArray(view.clansmen) ? view.clansmen : [];
--- a/apps/server/convex/getSnapshot.test.ts
+++ b/apps/server/convex/getSnapshot.test.ts
@@ -1,6 +1,7 @@
 import { describe, expect, it } from "vitest";
 import {
   deriveSeasonState,
+  getSnapshot,
   normalizePausedAtTs,
   resolveHeartbeatIntervalSecondsForSnap,
   resolvePauseStateForSnap,
@@ -12,6 +13,42 @@ import {
 const SEASON_LEN = 360;
 const WINTER_START = 110;
 const WINTER_DUR = 10;
+
+function createDb(tables: Record<string, any[]> = {}) {
+  return {
+    query(table: string) {
+      let rows = [...(tables[table] ?? [])];
+      const builder = {
+        withIndex(_name: string, apply: (q: any) => unknown) {
+          const clauses: Array<{ field: string; value: unknown }> = [];
+          const q = {
+            eq(field: string, value: unknown) {
+              clauses.push({ field, value });
+              return q;
+            },
+          };
+          apply(q);
+          rows = rows.filter((row) =>
+            clauses.every((clause) => row[clause.field] === clause.value),
+          );
+          return builder;
+        },
+        order(direction: "asc" | "desc") {
+          rows = [...rows].sort((a, b) =>
+            direction === "desc"
+              ? (b._creationTime ?? 0) - (a._creationTime ?? 0)
+              : (a._creationTime ?? 0) - (b._creationTime ?? 0),
+          );
+          return builder;
+        },
+        async first() {
+          return rows[0] ?? null;
+        },
+      };
+      return builder;
+    },
+  };
+}
 
 describe("resolveSeasonStateForSnap (issue #435 regression)", () => {
   it("season-freeze window: tick === seasonEndTick && !seasonFinalized returns persisted OLD season, NOT derived next season", () => {
@@ -196,5 +233,39 @@ describe("resolveHeartbeatIntervalSecondsForSnap", () => {
     expect(resolveHeartbeatIntervalSecondsForSnap({
       tickEpochDurationMs: 30_000,
     })).toBe(30);
+  });
+});
+
+describe("getSnapshot heartbeat interval", () => {
+  it("returns the latest persisted on-chain heartbeatIntervalSeconds", async () => {
+    const db = createDb({
+      worldSnapshot: [
+        {
+          _creationTime: 1,
+          tick: 8,
+          heartbeatIntervalSeconds: 60,
+          tickEpochStartedAt: 1_000,
+          tickEpochDurationMs: 60_000,
+          regions: [],
+          clans: [],
+          activeBanditId: 0,
+        },
+        {
+          _creationTime: 2,
+          tick: 8,
+          heartbeatIntervalSeconds: 45,
+          tickEpochStartedAt: 2_000,
+          tickEpochDurationMs: 60_000,
+          regions: [],
+          clans: [],
+          activeBanditId: 0,
+        },
+      ],
+    });
+
+    const result = await (getSnapshot as any)._handler({ db });
+
+    expect(result.heartbeatIntervalSeconds).toBe(45);
+    expect(result.tickEpoch.durationMs).toBe(60_000);
   });
 });
--- a/apps/server/convex/getSnapshot.test.ts
+++ b/apps/server/convex/getSnapshot.test.ts
@@ -2,6 +2,7 @@ import { describe, expect, it } from "vitest";
 import {
   deriveSeasonState,
   normalizePausedAtTs,
+  resolveHeartbeatIntervalSecondsForSnap,
   resolvePauseStateForSnap,
   resolveSeasonStateForSnap,
   type PersistedSnapshotSeasonFields,
@@ -180,5 +181,20 @@ describe("normalizePausedAtTs", () => {
       worldPaused: false,
       pausedAtTs: null,
     });
+  });
+});
+
+describe("resolveHeartbeatIntervalSecondsForSnap", () => {
+  it("prefers persisted on-chain heartbeatIntervalSeconds", () => {
+    expect(resolveHeartbeatIntervalSecondsForSnap({
+      heartbeatIntervalSeconds: 45,
+      tickEpochDurationMs: 60_000,
+    })).toBe(45);
+  });
+
+  it("falls back to legacy tickEpochDurationMs", () => {
+    expect(resolveHeartbeatIntervalSecondsForSnap({
+      tickEpochDurationMs: 30_000,
+    })).toBe(30);
   });
 });
--- a/apps/server/convex/getSnapshot.ts
+++ b/apps/server/convex/getSnapshot.ts
@@ -81,6 +81,7 @@ export function deriveSeasonState(tick: number): DerivedSeasonState {
  */
 export interface PersistedSnapshotSeasonFields {
   tick: number;
+  heartbeatIntervalSeconds?: number;
   currentSeasonNumber?: number;
   seasonStartTick?: number;
   seasonEndTick?: number;
@@ -141,13 +142,36 @@ export function resolvePauseStateForSnap(snap: {
   };
 }
 
+export function resolveHeartbeatIntervalSecondsForSnap(snap: {
+  heartbeatIntervalSeconds?: unknown;
+  tickEpochDurationMs?: unknown;
+}): number {
+  if (
+    typeof snap.heartbeatIntervalSeconds === "number" &&
+    Number.isFinite(snap.heartbeatIntervalSeconds) &&
+    snap.heartbeatIntervalSeconds > 0
+  ) {
+    return snap.heartbeatIntervalSeconds;
+  }
+  if (
+    typeof snap.tickEpochDurationMs === "number" &&
+    Number.isFinite(snap.tickEpochDurationMs) &&
+    snap.tickEpochDurationMs > 0
+  ) {
+    return Math.max(1, Math.floor(snap.tickEpochDurationMs / 1000));
+  }
+  return Number(HEARTBEAT_INTERVAL_SECONDS);
+}
+
 export const getSnapshot = query({
   handler: async (ctx) => {
     const snap = await ctx.db.query("worldSnapshot").order("desc").first();
+    const defaultHeartbeatIntervalSeconds = Number(HEARTBEAT_INTERVAL_SECONDS);
     if (!snap) {
       return {
         tick: 0,
-        tickEpoch: { startedAt: 0, durationMs: Number(HEARTBEAT_INTERVAL_SECONDS) * 1000 },
+        heartbeatIntervalSeconds: defaultHeartbeatIntervalSeconds,
+        tickEpoch: { startedAt: 0, durationMs: defaultHeartbeatIntervalSeconds * 1000 },
         regions: [],
         clans: [],
         activeBanditId: undefined,
@@ -185,6 +209,7 @@ export const getSnapshot = query({
         : null;
     return {
       tick: snap.tick,
+      heartbeatIntervalSeconds: resolveHeartbeatIntervalSecondsForSnap(snap),
       tickEpoch: {
         startedAt: snap.tickEpochStartedAt,
         durationMs: snap.tickEpochDurationMs,
--- a/apps/server/convex/getTickClock.ts
+++ b/apps/server/convex/getTickClock.ts
@@ -15,6 +15,7 @@ export const getTickClock = query({
       blockNumber: clock.blockNumber,
       tickEpochStartedAt: clock.tickEpochStartedAt,
       tickEpochDurationMs: clock.tickEpochDurationMs,
+      heartbeatIntervalSeconds: clock.heartbeatIntervalSeconds,
       currentSeasonNumber: clock.currentSeasonNumber,
       seasonStartTick: clock.seasonStartTick,
       seasonEndTick: clock.seasonEndTick,
--- a/apps/server/convex/heartbeat.ts
+++ b/apps/server/convex/heartbeat.ts
@@ -251,6 +251,10 @@ export const advanceTick = internalMutation({
     const baseTick = clockRow?.tick ?? snap.tick;
     const baseEpochStartedAt = clockRow?.tickEpochStartedAt ?? snap.tickEpochStartedAt;
     const baseEpochDurationMs = clockRow?.tickEpochDurationMs ?? snap.tickEpochDurationMs;
+    const baseHeartbeatIntervalSeconds =
+      clockRow?.heartbeatIntervalSeconds ??
+      snap.heartbeatIntervalSeconds ??
+      Math.max(1, Math.floor(baseEpochDurationMs / 1000));
 
     // Staleness gate: only advance if the current epoch has elapsed.
     const nowSeconds = Math.floor(Date.now() / 1000);
@@ -269,6 +273,7 @@ export const advanceTick = internalMutation({
         tick: newTick,
         tickEpochStartedAt: newEpochStartedAt,
         tickEpochDurationMs: baseEpochDurationMs,
+        heartbeatIntervalSeconds: baseHeartbeatIntervalSeconds,
         regions: snap.regions,
         clans: snap.clans,
       });
@@ -285,12 +290,14 @@ export const advanceTick = internalMutation({
         await ctx.db.patch(clockRow._id, {
           tick: newTick,
           tickEpochStartedAt: newEpochStartedAt,
+          heartbeatIntervalSeconds: baseHeartbeatIntervalSeconds,
         });
       } else {
         await ctx.db.insert("tickClock", {
           tick: newTick,
           tickEpochStartedAt: newEpochStartedAt,
           tickEpochDurationMs: baseEpochDurationMs,
+          heartbeatIntervalSeconds: baseHeartbeatIntervalSeconds,
           seasonStartTick: 0,
           seasonEndTick: 0,
           winterActive: false,
--- a/apps/server/convex/indexer.test.ts
+++ b/apps/server/convex/indexer.test.ts
@@ -76,6 +76,7 @@ type CommitSnapshotTestHandler = (
     snapshot: {
       blockNumber: number;
       txHash?: string;
+      heartbeatIntervalSeconds?: number;
       world: Record<string, unknown>;
       clans: unknown[];
     };
@@ -631,6 +632,48 @@ describe("legacy snapshot backfill", () => {
     expect(advancedClock.tick).toBe(13);
     expect(advancedClock.tickEpochStartedAt).toBe(1_060);
     expect(advancedClock.tickEpochDurationMs).toBe(60_000);
+
+    now.mockRestore();
+  });
+
+  it("persists on-chain heartbeatIntervalSeconds and patches interval-only changes", async () => {
+    const { db, tables } = createDb();
+    const now = vi.spyOn(Date, "now");
+    now.mockReturnValue(2_000_000);
+
+    await runCommitSnapshot(
+      { db },
+      {
+        snapshot: {
+          blockNumber: 200,
+          heartbeatIntervalSeconds: 45,
+          world: { currentTick: 20 },
+          clans: [],
+        },
+      },
+    );
+
+    expect(tables.tickClock?.[0]?.heartbeatIntervalSeconds).toBe(45);
+    expect(tables.tickClock?.[0]?.tickEpochDurationMs).toBe(45_000);
+    expect(tables.worldSnapshot?.[0]?.heartbeatIntervalSeconds).toBe(45);
+    expect(tables.worldSnapshot?.[0]?.tickEpochDurationMs).toBe(45_000);
+
+    await runCommitSnapshot(
+      { db },
+      {
+        snapshot: {
+          blockNumber: 201,
+          heartbeatIntervalSeconds: 30,
+          world: { currentTick: 20 },
+          clans: [],
+        },
+      },
+    );
+
+    expect(tables.tickClock?.[0]?.heartbeatIntervalSeconds).toBe(30);
+    expect(tables.tickClock?.[0]?.tickEpochDurationMs).toBe(30_000);
+    expect(tables.worldSnapshot?.[0]?.heartbeatIntervalSeconds).toBe(30);
+    expect(tables.worldSnapshot?.[0]?.tickEpochDurationMs).toBe(30_000);
 
     now.mockRestore();
   });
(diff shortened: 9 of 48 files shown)


mcorrig4:
@codex please review

claude-do:
# Phase Super-Swarm Synthesis — PR #503 (head `a26534e`)

**Models run:** Codex 5.3 ✓ | Codex 5.4 ✓ | Codex 5.5 ✓ | Opus 4.6 ✓ | Opus 4.7 ✓ | Gemini 3.1 Pro ✓
**PR:** [#503 feat(heartbeat): on-chain interval as single source of truth + self-scheduling runner](https://github.com/clan-world/clan-world-game/pull/503)
**Diff:** 44 files, +990 / -303 lines

## Summary

**Verdict: NEEDS_FIXES — 2 HIGHs + 4 cross-cutting MEDs.** Architectural shape is correct and unanimously praised: on-chain as source of truth, self-scheduling, hot-loop guard, indexer persistence, Telegram resilience. Two real bugs need fixing before merge — one false-positive-alert path that 5/6 reviewers caught, and a settle-latch regression on the standalone launcher path that codex 5.4 caught alone. Plus 4 multi-reviewer MEDs around alert deduping + RPC outage handling worth addressing in the same fix-round.

## MUST FIX (cross-model convergence)

| File:line | Models | Severity | Finding | Fix |
|---|---|---|---|---|
| `packages/runner/src/heartbeatScheduler.ts:155-168, 181-215` + `runnerCastHeartbeat.ts:111-128` | codex 5.3, 5.4, 5.5, gemini, opus 4.7 (**5/6**) | HIGH | `HeartbeatRateLimitedError` + abort signal both fall into the generic 5-attempt retry loop and fire false Telegram alerts. Pre-refactor rate-limited was caught + continued (no alert); now it runs 1+2+5+10s of wasted retries then pages do-crew. Same path means SIGTERM during deploy alerts "heartbeat failed: aborted". | Early-exit `attemptHeartbeatWithBackoff`: if `err instanceof HeartbeatRateLimitedError`, return without retrying or alerting (loop re-reads nextHeartbeatAtTs naturally). If `signal.aborted`, return silently without alert. Add tests for both paths. |
| `scripts/start-heartbeat-loop.sh` + `packages/runner/src/heartbeatLoopMain.ts:18-23` | codex 5.4 (1/6 but verified) | HIGH | Standalone launcher invokes scheduler WITHOUT `settleLatch`. Full runner wires it correctly, but documented bash launcher loses Cycle B coordination — reintroduces the "heartbeat cuts off order-submission window" bug the latch was designed to prevent. | Either: (a) wire a settle-latch (or a no-op equivalent) into `heartbeatLoopMain.ts` so standalone has Cycle B awareness, OR (b) document standalone is for emergencies only and full runner is the production path. Pick (a) — making the script's behavior match expectations is safer than hoping operators read the runbook. |

## SHOULD FIX (multi-reviewer MED)

| File:line | Models | Finding | Fix |
|---|---|---|---|
| `heartbeatScheduler.ts:155-168` outer alert path | gemini MED, opus 4.7 M1 | No dedup window on Telegram alerts. During sustained RPC outage, ~150 alerts/hour. | Track `lastAlertAtMs`, suppress new alerts within 5-10 min of previous OR "don't alert again until next successful heartbeat." |
| `heartbeatScheduler.ts:115` `readNextHeartbeatAtTs` failure path | gemini MED, opus 4.6 M1 | RPC failure → always 1s sleep, no escalation → hammers RPC every ~1s during outage. | Track consecutive read failures, escalate via same `HEARTBEAT_RETRY_BACKOFF_MS` sequence, reset on success. |
| `runnerCastHeartbeat.ts:96-118` `HeartbeatTimeoutError` path | opus 4.7 M4 | Receipt-wait timeout doesn't mean tx didn't land. Retry submits SECOND heartbeat() → wasted gas + may revert with rate-limited (→ cascades with HIGH #1). | Before retry after `HeartbeatTimeoutError`, re-read `nextHeartbeatAtTs`. If advanced, the tx landed — treat as success. |
| `heartbeatScheduler.ts:91-101` boot interval-read failure | opus 4.7 M5 | Writes `lastFireResult: 'error'` with "boot interval read failed" — but no fire was attempted. Conflates failure classes. | Add `'boot-error'` literal to runnerStatus enum, OR separate `lastBootError` field. |

## DEFER (single-reviewer MED/LOW — follow-up issues, not blockers)

| Finding | Models | Plan |
|---|---|---|
| Deleted slow-callHeartbeat regression test | opus 4.7 M3 | File backlog issue. Re-add an equivalent test pinning the settledSnapshot-before-callHeartbeat invariant. |
| Retry-exhaustion status not distinct from single failure | opus 4.7 M2 | Add `lastExhaustionAt` field or prefix `lastFailureMessage` with "exhausted N retries:". |
| Scheduler `.catch` doesn't exit/restart on crash | opus 4.6 M2 | Have `.catch` call `process.exit(1)` so tmux wrapper restarts. Low probability given defensive coding. |
| runnerStatus query has no `.take(50)` cap | opus 4.7 M6 | Add limit before more runner types appear. |
| HOT_LOOP_GUARD_MS (1000) + jitter (500) overlap = 1.5s instead of 1s | gemini LOW | Cosmetic. Either accept 1.5s or skip jitter when hot-loop guard fires. |
| Missing rate-limit test | gemini LOW | Will be added as part of MUST FIX #1's test coverage. |

## Cross-model overlap stats

- **Findings flagged by 3+ models**: 2 (rate-limit/abort → false alert + retries; RPC-read-failure spam)
- **Findings flagged by 2 models**: 1 (Telegram alert dedup)
- **Single-model findings**: ~7
- **HIGH or code-bug findings**: 2 (consensus + 1 verified-real-regression from codex 5.4)
- **CLEAN verdicts**: 0 unanimous, but opus 4.6 said "merge-ready with operational advisories"

## Per-model verdicts

- **Codex 5.3:** NEEDS_FIXES — 1 HIGH (abort alert) + several MEDs
- **Codex 5.4:** NEEDS_FIXES — 1 HIGH (settle-latch standalone regression) + 1 MED (abort alert)
- **Codex 5.5:** NEEDS_FIXES — abort path MED + minor tightening
- **Opus 4.6:** CLEAN with 2 operational MEDs (RPC backoff + scheduler crash hang)
- **Opus 4.7:** NEEDS_FIXES — 1 HIGH (HeartbeatRateLimitedError) + 6 MEDs (most thorough review)
- **Gemini 3.1 Pro:** NEEDS_FIXES — 1 HIGH (spurious alerts) + 2 MEDs

## Merge recommendation

**HOLD until fix-round addresses MUST FIX items.** Dispatching codex 5-stage fix-round (or extending the existing session) to address the 2 HIGHs + 4 SHOULD-FIX MEDs. The DEFER list goes to follow-up issues — not gating merge.

claude-do:
## Cloud-review fix-round addressed (commit `ab3bd4f`)

Following the super-swarm fix-round (`2dd279b`), this round addresses cloud reviewer comments on the original `a26534e`:

### Addressed
- ✅ **gemini HIGH** ClanWorld.sol:5169 — `heartbeatIntervalSeconds()` pure→view, reads `LibStorage.appStorage().heartbeatIntervalSeconds` via `LibWorldClock`. Genesis fallback to compile-time constant preserved.
- ✅ **gemini MED** clansmen.ts:147 — now uses `resolveHeartbeatIntervalSecondsForSnap` helper
- ✅ **Copilot MED** TopHud.tsx:73 — anchor refresh on `fallbackTickDurationMs` change + new `TopHud.test.ts`
- ✅ **Copilot MED** heartbeatScheduler.ts:187 — `lastFireAt` + alert dedup use `deps.nowMs()` override
- ✅ **Copilot MED (SECURITY)** runnerStatus.ts:24 — `getRunnerStatus` now requires `INDEXER_SECRET` via `requireIndexerSecret` gate; new test verifies unauthenticated rejection

### Verified duplicates (already in super-swarm fix-round `2dd279b`)
- codex P2 — HeartbeatRateLimitedError retry escalation → already fixed
- codex P3 — abort-during-retry alert suppression → already fixed

### Test verdict
- `pnpm --filter @clan-world/server test`: **81 passed** (was 79)
- `pnpm --filter @clan-world/runner test`: **137 passed** (was 136)
- All typechecks green
- ABI regen clean
- `forge build src/ClanWorld.sol`: PASS

claude-do:
# R2 Super-Swarm Synthesis — PR #503 (head `ab3bd4f`)

**Models run:** Codex 5.3 ✓ | Codex 5.4 ✓ | Codex 5.5 ✓ | Opus 4.6 ✓ | Opus 4.7 ✓ | Gemini 3.1 Pro ✓
**Goal:** Verify R1 + cloud-review fixes landed correctly. Find new issues introduced by the fixes themselves.

## Summary

**Verdict: NEEDS_FIXES (small) — 2 MEDs converge across reviewers, 1 architectural MED from opus 4.7 alone but real, 1 latent deadlock from codex 5.4 + opus 4.6.** R1 + cloud fixes unanimously VERIFIED FIXED with file:line refs across 11 findings. The fix-round itself introduced one over-broad guard (alert dedup) and a Convex-patch semantics oversight (stale lastFailureMessage). Plus opus 4.7 caught a webhook-POST regression vs architecture-decisions.md.

## MUST FIX (cross-model convergence)

### R2-M1 — Telegram alert dedup over-broad (4/6 reviewers: codex 5.3, 5.4, 5.5, gemini)

`packages/runner/src/heartbeatScheduler.ts:165-170`. The 5-min dedup window suppresses alerts based ONLY on `lastAlertAtMs` (single timestamp). If `rpc-down` exhausts retries at t=0 and a DIFFERENT failure (e.g., `revert`, `timeout`, contract-paused) lands at t=3min, the second alert is dropped — exactly the failure mode oncall most needs visibility on.

**Fix:** key the dedup by error class/message-fingerprint, not a single timestamp. Either:
- (a) `Map<lastFireResult, lastAlertAtMs>` — per-class dedup window
- (b) Suppress only IDENTICAL messages within window
- (c) Always alert on RESULT-CLASS transitions; dedup only on same-class repeats

## SHOULD FIX

### R2-M2 — Stale `lastFailureMessage` persists after success (opus 4.6 MED + opus 4.7 LOW, 2/6 cross-confirmed)

`apps/server/convex/runnerStatus.ts:50-57`. On success, scheduler posts `lastFailureMessage: undefined`. Convex's `ctx.db.patch()` SILENTLY IGNORES `undefined` values — so the prior failure message persists alongside `lastFireResult: 'success'`. Opus 4.6 notes the inconsistency: `lastFireAt` uses conditional spread (showing awareness) but `lastFailureMessage` doesn't.

**Fix:** explicitly clear by passing `""` on success (and widen validator if needed), OR use conditional spread to omit when clearing AND patch with explicit empty string on success path.

### R2-M3 — Webhook POST removed from standalone keeper (opus 4.7 MED, 1/6 but architectural)

`scripts/start-heartbeat-loop.sh` previously POSTed `/api/heartbeat-webhook` after each `cast send`. The new TS launcher (`heartbeatLoopMain.ts` → `RunnerCastHeartbeat`) drops that POST entirely. Per `docs/reference/architecture-decisions.md` the indexer trigger is **webhook-primary, 5s poll = safety net**. The standalone loop now relies entirely on the safety net.

**Fix options:**
- (a) Re-add the webhook POST inside `RunnerCastHeartbeat.callHeartbeat()` after `waitForTransactionReceipt`
- (b) Add it in `heartbeatLoopMain.ts` post-fire hook
- (c) Document the architectural shift in a runbook + ADR update (only if intentional)

Pick (a) — restores parity with old keeper without touching the scheduler's contract.

### R2-M4 — Standalone deadlock in static-tick / stub Convex (codex 5.4 + opus 4.6 LOW, 2/6 cross-confirmed)

`convexSnapshotSettleLatch.ts` waits for `worldSnapshot.tick` to advance. If `CLANWORLD_USE_STUB_CONVEX=true` or Convex isn't reachable, the stub returns `tick: 0` forever. After first heartbeat sets `lastHeartbeatForTick = 0`, the scheduler waits indefinitely at the settle gate. Codex 5.4 verified `pollChainTick.ts:8` documents the stub-mode `tick: 0` invariant.

**Fix:**
- Detect stub mode at boot (via `CONVEX_URL` env presence) and skip the settle-latch entirely OR replace with a permissive no-op latch
- Log a startup warning when stub Convex is detected with a settle-latch active

## DEFER (single-reviewer LOWs — follow-up issues)

| Finding | Models | Plan |
|---|---|---|
| `Date.now()` in `runnerCastHeartbeat.ts:98,110` not threaded `nowMs()` | gemini | Single follow-up issue. Caller adapter time-source could matter for rate-limit-time-check tests. |
| `sleepWithSignal` duplicated across 3 files | opus 4.6 L1 | Extract to `packages/runner/src/util.ts`. Tiny cleanup. |
| `classifyHeartbeatError` has dead `'rate-limited'` branch | opus 4.7 L1 | `HeartbeatRateLimitedError` is caught explicitly before this function runs. Delete the unreachable branch. |
| Convex latch `setTimeout` chain no backoff on `getSnapshot` failure | opus 4.7 cross-cutting | Mirror main scheduler's escalating backoff. |
| No `forge test` for `ClanWorld.heartbeatIntervalSeconds()` | opus 4.7 L4 | 5-line forge test would close the loop. Hackathon-acceptable to defer. |
| `getRunnerStatus` test coverage thin (no happy-path) | opus 4.7 L5 + codex 5.5 LOW | Add 1 happy-path test with valid INDEXER_SECRET. |
| `heartbeatIntervalSeconds` boot-once → `runnerStatus.heartbeatIntervalSeconds` stale | opus 4.7 L6 | Already documented in runner README; operational concern only. |

## R1 + cloud fixes verification (unanimous 6/6 VERIFIED FIXED)

Every R1 + cloud-review finding was independently confirmed by all 6 reviewers with file:line evidence. No R1 regressions, no R1-fix hallucinations. Highest-quality R2 outcome we've seen on this repo.

## Cross-model overlap stats

- **Findings flagged by 4+ models**: 1 (alert dedup over-broad)
- **Findings flagged by 2 models**: 2 (lastFailureMessage stale; standalone deadlock)
- **Single-model significant findings**: 1 (webhook POST regression — opus 4.7 only, but architecturally significant per ADR)
- **HIGH severity findings**: 0
- **NEW MEDs**: 4

## Per-model verdicts

- **Codex 5.3:** NEEDS_FIXES — 1 MED (alert dedup)
- **Codex 5.4:** NEEDS_FIXES — 2 MEDs (alert dedup + standalone deadlock)
- **Codex 5.5:** NEEDS_FIXES — 1 MED (alert dedup) + 3 test-coverage LOWs
- **Opus 4.6:** NEEDS_FIXES — 1 MED (lastFailureMessage) + 3 LOWs
- **Opus 4.7:** NEEDS_FIXES — 1 MED (webhook POST) + 6 LOWs
- **Gemini 3.1 Pro:** NEEDS_FIXES — 1 MED (alert dedup) + 3 LOWs

## Merge recommendation

**HOLD until R3 fix-round addresses 4 MEDs.** Then re-verify with a tighter sanity-pass (no full super-swarm needed since R1+cloud convergence is already validated).

claude-do:
# R3 Super-Swarm Synthesis — PR #503 (head `d3df280`)

**Models run:** Codex 5.3 ✓ | Codex 5.4 ✓ | Codex 5.5 ✓ | Opus 4.6 ✓ | Opus 4.7 ✓ | Gemini 3.1 Pro ✓
**Goal:** Verify R2 fix-round, find issues introduced BY the R2 fixes.

## Summary

**Verdict: NEEDS_FIXES — 2 HIGHs + 3 MEDs introduced by R2 fix-round.** R2's 4 MEDs are all verified fixed by 6/6 reviewers. But the R2 webhook POST fix (R2-M3) introduced 2 new bugs that 4-of-6 reviewers caught, and gemini caught a critical rate-limited deadlock that orchestrator verified by reading actual code. Codex 5.5 also caught a backwards-compat env var regression. Liam's caution about "too many R2 surprises to assume green" — vindicated.

## MUST FIX

### R3-H1 — Rate-limited heartbeat causes settle-latch deadlock (gemini, VERIFIED by orchestrator)

`packages/runner/src/heartbeatScheduler.ts:149-150`:
```ts
if (result.success) {
  if (deps.settleLatch) lastHeartbeatForTick = settledSnapshot;
  if (!result.rateLimited) lastAlertAtMs.clear();
}
```

`HeartbeatRateLimitedError` returns `{success: true, rateLimited: true}` (R2-M1 fix). The success branch correctly skips alert-clear for rate-limited, BUT STILL updates `lastHeartbeatForTick = settledSnapshot`. Since rate-limited heartbeats aren't actually mined, the world tick doesn't advance — but `lastHeartbeatForTick` got bumped. Next iteration the latch check `settledSnapshot <= lastHeartbeatForTick` blocks forever.

**Orchestrator verified by reading `heartbeatScheduler.ts:149-150` directly — this is a real bug. Gemini caught it alone (other reviewers focused on R2 verification + missed the conditional).**

**Fix:** wrap the `lastHeartbeatForTick = settledSnapshot` in `if (!result.rateLimited)` (the line below it already gates correctly):
```ts
if (result.success) {
  if (deps.settleLatch && !result.rateLimited) lastHeartbeatForTick = settledSnapshot;
  if (!result.rateLimited) lastAlertAtMs.clear();
}
```
OR fold into one guard:
```ts
if (result.success && !result.rateLimited) {
  if (deps.settleLatch) lastHeartbeatForTick = settledSnapshot;
  lastAlertAtMs.clear();
}
```

Add test: rate-limited heartbeat → settledSnapshot stays same → next iteration's settle check doesn't deadlock.

### R3-H2 — Webhook POST URL parse outside try-catch (3/6: codex 5.3, 5.4, 5.5)

`packages/runner/src/runnerCastHeartbeat.ts:169-170` (and surrounding):
```ts
const webhookUrl = new URL('/api/heartbeat-webhook', this.convexWebhookUrl);  // <-- outside try
try {
  // fetch...
}
```

If `CONVEX_WEBHOOK_URL` is malformed (or non-URL-shaped placeholder string), `new URL()` throws SYNCHRONOUSLY. Since it's BEFORE the try block, the throw propagates out of `postHeartbeatWebhook` → out of `callHeartbeat` → and a confirmed on-chain heartbeat tx gets reported as a FAILURE to the scheduler. That triggers retries, alerts, etc. for a tx that actually succeeded.

**Fix:** move the `new URL(...)` INSIDE the try block, OR validate the env var once at construction time. Best-effort means truly best-effort.

## SHOULD FIX

### R3-M1 — Webhook POST `fetch` has no timeout (4/6: codex 5.3 MED, 5.4 MED, 5.5 HIGH, opus 4.7 MED)

`runnerCastHeartbeat.ts:172` `await fetch(webhookUrl, ...)` has no `AbortSignal`/timeout. A slow or hung Convex webhook endpoint blocks `callHeartbeat()` indefinitely after the on-chain receipt is confirmed. Scheduler can't progress → missed heartbeats.

**Fix:** add `signal: AbortSignal.timeout(5_000)` to the fetch options. Best-effort side effect cannot hold the main loop hostage.

### R3-M2 — Settle latch deadlocks when Convex is configured-but-unreachable (4/6: codex 5.3 LOW, 5.4 MED, 5.5 MED, opus 4.7 MED)

`packages/runner/src/convexSnapshotSettleLatch.ts:20-31`. The `isStub` fix (R2-M4) covers stub clients. But a `RealConvexClient` pointed at a dead/misconfigured Convex deployment hits the SAME deadlock: `lastSeenTick` stays `-1`, scheduler blocks at `settledSnapshot (-1) <= lastHeartbeatForTick (-1)` before the first fire.

**Fix options:**
- (a) After N consecutive `getSnapshot` failures, fail-open by advancing `lastSeenTick` past `lastHeartbeatForTick` (logged)
- (b) Detect first-fire-readiness: if `lastSeenTick === -1` AND we haven't fired yet, allow the first heartbeat to proceed
- (c) Boot-time Convex liveness probe with timeout; bail out (or skip latch) if unreachable

Pick (b) — least invasive, handles the first-fire boot case which is the deadlock window.

### R3-M3 — Backwards-compat env var regression (codex 5.5)

`packages/runner/src/runnerCastHeartbeat.ts:56` (and similar). Old `scripts/start-heartbeat-loop.sh` derived `CONVEX_WEBHOOK_URL` from `CONVEX_DEPLOY_URL` via `${CONVEX_DEPLOY_URL/.convex.cloud/.convex.site}`. New TS launcher reads ONLY `CONVEX_WEBHOOK_URL`. Existing VPS setups (per `docs/runbooks/base-sepolia-deployment.md`) will silently skip webhook posts.

**Fix:** in `RunnerCastHeartbeat` config, if `CONVEX_WEBHOOK_URL` is unset but `CONVEX_DEPLOY_URL` is set, derive `.convex.site` automatically. Log a deprecation warning suggesting explicit env var. OR update all runbooks + `.env.template` to require explicit `CONVEX_WEBHOOK_URL` (fail loud if unset on prod path).

Pick the derive-with-warning approach — preserves existing ops envs while documenting the migration.

## DEFER (LOWs — follow-up issues)

| Finding | Models | Plan |
|---|---|---|
| `sendTelegramAlert` no timeout | opus 4.7 LOW-1 | Apply same `AbortSignal.timeout(5_000)` pattern. Tiny, ship with R3 fix-round. |
| `lastAlertAtMs.set()` runs before alert send confirmed | opus 4.7 LOW-2 | Subtle — silent dedup if Telegram failed. Move `.set()` after `alertResult.ok`. |
| `classifyHeartbeatError` dead `rate-limited` regex branch | opus 4.6 + 4.7 | Cleanup. Defer. |
| `sleepWithSignal` duplicated 3x | opus 4.6 | Extract to util. Defer. |
| `isHeartbeatTimeoutError` name match too broad | opus 4.7 LOW-5 | Mitigated by `nextAfterTimeout > nextHeartbeatAtTs` check. Acceptable. |
| Cold-start race on `lastSeenTick = -1` | opus 4.7 LOW-3 | Resolves itself within 1s. Cosmetic. |
| R3-M3 fix needs better test coverage (invalid URL, non-OK, hung fetch) | codex 5.4 MED | Bundle test additions with R3-H2/R3-M1 fixes. |

## R2 fixes verification (6/6 VERIFIED FIXED)

All 4 R2 MEDs unanimously verified with file:line evidence by all 6 reviewers. No R2 regressions, no R2-fix hallucinations.

## Cross-model overlap stats

- **HIGHs caught by 1 model + verified by orchestrator**: 1 (rate-limited deadlock — gemini)
- **HIGHs caught by 3+ models**: 1 (webhook URL outside try-catch)
- **MEDs caught by 4+ models**: 2 (webhook fetch no timeout; real-unreachable Convex deadlock)
- **MEDs caught by 1 model**: 1 (backwards-compat env var — codex 5.5)
- **R2 fixes verified by all 6**: 4/4

## Per-model verdicts

- **Codex 5.3:** NEEDS_FIXES — 2 MEDs (webhook URL + fetch timeout) + 1 LOW
- **Codex 5.4:** NEEDS_FIXES — 1 HIGH (webhook URL outside try) + 3 MEDs
- **Codex 5.5:** NEEDS_FIXES — 1 HIGH (webhook URL + timeout combined) + 2 MEDs
- **Opus 4.6:** CLEAN with 3 LOWs (less critical than other reviewers)
- **Opus 4.7:** NEEDS_FIXES — 2 MEDs + 5 LOWs (matched codex on webhook concerns)
- **Gemini 3.1 Pro:** NEEDS_FIXES — 2 HIGHs (rate-limited deadlock + webhook timeout) — caught rate-limited deadlock alone

## Cross-model disagreement resolution

Opus 4.6 verdict was "CLEAN with LOWs only" while 5 other reviewers flagged the webhook bugs as MED/HIGH. Per memory `feedback_super_swarm_cross_model_disagreement_resolution`: trust convergence pattern + verify actual code. The webhook URL outside try-catch is verifiable; the timeout absence is verifiable; the rate-limited deadlock is verifiable. Opus 4.6 was less critical, NOT correct.

## Merge recommendation

**HOLD until R3 fix-round addresses 2 HIGHs + 3 MEDs.** Then a R4 sanity pass (orchestrator's call: full super-swarm vs tight local-swarm — Liam's prior directive favored full given the cascading-fix pattern).

claude-do:
# R4 Super-Swarm Synthesis — PR #503 (head `bf26ea8`)

**Models run:** Codex 5.3 ✓ | Codex 5.4 ✓ | Codex 5.5 ✓ | Opus 4.6 ✓ | Opus 4.7 ✓ | Gemini 3.1 Pro ✗ (empty-response error)
**Goal:** Verify R3 fix-round, find issues introduced by R3.

## Summary

**Verdict: NEEDS_FIXES (small, scoped) — 2 NEW MEDs converge across reviewers, plus a disputed settle-latch HIGH (2/5 codex flag as HIGH, opus 4.7 verifies as mitigated).** All 7 R3 fixes unanimously VERIFIED FIXED across 5 reviewers with file:line refs. The 4-round dedup-fix-verify cycle is converging cleanly.

## DISPUTED — Cross-model disagreement (settle-latch first-fire watermark)

**codex 5.4 + 5.5 flag as HIGH:** `heartbeatScheduler.ts:149-150` — first fire stores `lastHeartbeatForTick = Math.max(0, settledSnapshot)` = `0` when Convex was unreachable at first fire. Subsequent iterations only require `settledSnapshot > 0`, which can be satisfied by Convex reporting an OLD tick (e.g. snapshot lagging by 30s) — gate clears even though the heartbeat just fired hasn't settled.

**opus 4.7 walked the same code path and verified as mitigated:**
> "`Math.max(0, settledSnapshot)` walked through end-to-end for the cold-start case (settledSnapshot=-1 → lastHeartbeatForTick=0 → requires next settle ≥1). Holds as long as the heartbeat actually advances the chain (which is its job). If the very first tx is dropped/lost the watermark advances but Cycle B never observes a new tick → next iteration would wait. Mitigated by R3-H2/M1 retry path + alerting, but worth a deliberate comment."

**Orchestrator interpretation:** Both analyses are correct for different failure modes. The codex finding is real for degraded-Convex scenarios where Convex lags behind chain. Opus 4.7 is right that hackathon-scope normal-operation handles it. **Recommend deferring as a "watermark hardening" follow-up issue rather than blocking merge** — the deadlock R3 was solving is the more critical bug. Adding a deliberate comment explaining the trade-off is sufficient for now.

## MUST FIX (cross-model convergence)

### R4-M1 — `deriveConvexWebhookUrl` silently mis-derives non-`.convex.cloud` URLs (3/5: codex 5.4, 5.5, opus 4.7)

`packages/runner/src/runnerCastHeartbeat.ts:227-230` blind `.replace('.convex.cloud', '.convex.site')`. For self-hosted Convex, custom domains, or staging aliases:
- Replace is a no-op → derived webhook URL points back at the deploy URL
- Or replace mis-derives → wrong host
- Either way, deprecation warning fires saying "set explicitly" but runner posts every heartbeat to the wrong place (best-effort caught, but logs spam forever)

**Fix:** add a one-line guard at the top of `deriveConvexWebhookUrl`:
```ts
if (!convexDeployUrl.includes('.convex.cloud')) return undefined;
```

Only derive when the suffix actually matches. For non-`.convex.cloud` URLs, leave webhook disabled with the existing deprecation message.

Add test: non-`.convex.cloud` URL → returns `undefined` → no webhook POST.

### R4-M2 — Telegram-failure path overwrites `lastFailureMessage` with Telegram error (opus 4.7)

`packages/runner/src/heartbeatScheduler.ts:183-191` — when `sendTelegramAlert` returns `{ok: false}`, the code posts `runnerStatus { lastFireResult: 'error', lastFailureMessage: 'Telegram alert failed: …' }`. This OVERWRITES the heartbeat-failure message that was just posted inside `attemptHeartbeatWithBackoff`. Ops dashboards looking at latest `runnerStatus` see "Telegram alert failed" with no clue what actually broke the heartbeat.

**Fix options:**
- (a) Add a separate `'alert-error'` literal to the runnerStatus enum; post under that class instead of clobbering `'error'`
- (b) Concatenate: `lastFailureMessage: 'heartbeat: <X>; alert: <Y>'`
- (c) Don't write runnerStatus on alert failure — the heartbeat-failure status is already there

Pick (c) — simplest. Just log the alert failure to console; don't touch runnerStatus. The heartbeat failure message stays visible. Or (a) if we want explicit observability.

## DEFER (LOWs — file as follow-up issues)

| Finding | Models | Plan |
|---|---|---|
| codex 5.3 LOW: alert noise during Telegram outages (intentional tradeoff) | codex 5.3 + codex 5.5 LOW | Already noted as intentional; document in runner README. |
| opus 4.7 L1: rate-limited success leaves dedup map intact | opus 4.7 | Cosmetic asymmetry. Add a comment. |
| opus 4.7 L2: TopHud anchor reset on duration change | opus 4.7 | Cosmetic UI nit. |
| opus 4.7 L3: Settle-latch wait re-reads `nextHeartbeatAtTs` every 1s | opus 4.7 | 30 RPC calls per 30s wait. Cache. |
| opus 4.7 L4: `(getSnapshot as any)._handler({db})` test pattern | opus 4.7 | Undocumented internal of Convex query. |
| opus 4.7 L5: Empty `CONVEX_WEBHOOK_URL=""` falls through to derivation | opus 4.7 | Minor. Treat empty as set. |
| opus 4.7 L6: Hot-loop guard assumes single-cause failure | opus 4.7 | Real anomaly that would page anyway. |
| settle-latch watermark hardening (DISPUTED HIGH above) | codex 5.4 + 5.5 | Defer as separate follow-up issue with detailed scenarios + opus 4.7's "deliberate comment" suggestion implemented inline. |

## R3 fixes verification (5/5 VERIFIED FIXED — gemini failed)

All 7 R3 findings independently VERIFIED FIXED by 5 reviewers with file:line evidence.

## Cross-model overlap stats

- **Findings flagged by 3+ models**: 1 (URL derive M1)
- **Findings flagged by 2 models**: 1 (settle-latch DISPUTED)
- **Single-model significant findings**: 1 (Telegram overwrite M2)
- **HIGH severity findings (unanimous)**: 0
- **NEW MEDs (with consensus)**: 2

## Per-model verdicts

- **Codex 5.3:** CLEAN — 1 LOW (telegram outage tradeoff)
- **Codex 5.4:** NEEDS_FIXES — 1 HIGH (settle-latch, disputed) + 1 MED (URL derive)
- **Codex 5.5:** NEEDS_FIXES — 1 HIGH (settle-latch, disputed) + 1 MED (URL derive) + 1 LOW
- **Opus 4.6:** CLEAN
- **Opus 4.7:** CLEAN/MERGE — 2 MEDs (URL derive + Telegram overwrite) + 6 LOWs
- **Gemini 3.1 Pro:** ERROR — empty-response (Antigravity glitch; not counted)

## Merge recommendation

**Two paths forward:**

(a) **Quick fix-round** — apply R4-M1 (1-line guard on URL derive) + R4-M2 (option (c): drop runnerStatus write on alert failure), file the settle-latch HIGH as a deferred hardening issue with a deliberate comment. Then R5 sanity-pass + merge. Total ~15-20 min wall-clock.

(b) **Merge as-is** — file all 3 (R4-M1, R4-M2, settle-latch) as follow-up issues. The R3 fix-round resolved the actual deadlocks. The remaining concerns are operational polish.

Recommend (a) — the 2 MEDs are small, scoped, and worth catching. The settle-latch nuance gets a comment + follow-up.

claude-do:
# R5 Super-Swarm Synthesis — PR #503 (head `5227f80`)

**Models run:** Codex 5.3 ✓ | Codex 5.4 ✓ | Codex 5.5 ✓ | Opus 4.6 ✓ | Opus 4.7 ✓ | Gemini 3.1 Pro ✓ (recovered!)
**Goal:** Verify R4 fix-round.

## Summary

**Verdict: NEAR-CLEAN — 1 cross-model split MED (codex MED, opus/gemini LOW), 1 single-model MED, 3 LOWs.** R4 fixes 6/6 unanimously VERIFIED FIXED. The remaining concerns are operational polish + a tiny security-edge-case hardening. 5-line fix to close everything; recommend one more tight round then merge.

## Cross-model severity disagreement

`deriveConvexWebhookUrl` uses `.includes('.convex.cloud')` substring match:
- **Codex 5.3, 5.4, 5.5 (all 3 codex):** MED — `https://attacker.convex.cloud.evil.com` matches and rewrites to `https://attacker.convex.site.evil.com`. Operator-controlled env var but still a one-line hardening worth applying.
- **Opus 4.6, 4.7, Gemini (all 3 non-codex):** LOW — theoretical, operator-controlled, best-effort path. Acceptable.

**Orchestrator resolution:** Per memory `feedback_super_swarm_cross_model_disagreement_resolution`, verify against actual code AND consider the cost. The fix is 5 lines — `new URL().hostname.endsWith('.convex.cloud')`. The cost is trivial; the benefit is defense-in-depth. **Apply the fix.**

## SHOULD FIX

### R5-M1 — `deriveConvexWebhookUrl` hostname check (3/3 codex MED, 3/3 non-codex LOW — 6/6 flagged)

`packages/runner/src/runnerCastHeartbeat.ts:240-244`. Change from substring to URL-parse + hostname-suffix:
```ts
function deriveConvexWebhookUrl(convexDeployUrl: string): string | undefined {
  try {
    const url = new URL(convexDeployUrl);
    if (!url.hostname.endsWith('.convex.cloud')) return undefined;
    url.hostname = url.hostname.replace(/\.convex\.cloud$/, '.convex.site');
    return url.toString().replace(/\/$/, '');
  } catch {
    return undefined;
  }
}
```

Add test: `https://attacker.convex.cloud.evil.com` → returns `undefined`; `https://foo.convex.cloud` → returns `https://foo.convex.site`.

### R5-M2 — `.env.template` has `CONVEX_WEBHOOK_URL=` empty (codex 5.5 alone, but real)

`.env.template:49`. With R4-L5 fix, empty-present = "webhook intentionally disabled". So users who copy `.env.template` and only set `CONVEX_DEPLOY_URL` (the normal path) get webhook **disabled** instead of derived — defeating R4-M3 backwards-compat.

**Fix:** comment out the empty line so it's unset by default:
```env
# CONVEX_WEBHOOK_URL=  # (uncomment + set to override the .convex.cloud → .convex.site derivation)
```

Add test: env without `CONVEX_WEBHOOK_URL` key → derivation works as expected.

## DEFER / LOW

| Finding | Models | Plan |
|---|---|---|
| Settle-latch comment missing `#511` reference | codex 5.3, 5.4, 5.5, opus 4.6, opus 4.7 (5/6 LOW) | 1-character fix: `See follow-up.` → `See #511.`. Apply with R5-M1/M2 since we're in the file. |
| Telegram alert delivery failure not visible in runnerStatus | codex 5.4, 5.5, opus 4.7 LOW | Acceptable for hackathon (R4-M2 fix was correct — alert-fail shouldn't clobber heartbeat-fail). Add `lastAlertDeliveryAt`/`lastAlertDeliveryError` field if/when ops need it. File follow-up. |
| `sleepWithSignal` duplicated 3x | opus 4.6 LOW | Cosmetic. Extract to util later. |

## R4 fixes verification (6/6 VERIFIED FIXED)

All 4 R4 targets unanimously verified across 6 reviewers. R4-disputed-HIGH (settle-latch) comment exists; only missing the `#511` literal.

## Cross-model overlap stats

- **NEW MEDs (codex view)**: 2 (URL substring + `.env.template`)
- **NEW MEDs (opus/gemini view)**: 0
- **Unanimous LOWs**: 1 (comment #511 ref)
- **HIGH findings**: 0
- **R4 fixes verified**: 4/4 unanimous

## Per-model verdicts

- **Codex 5.3:** NEEDS_FIXES — 1 MED (URL substring) + 1 LOW (comment)
- **Codex 5.4:** NEEDS_FIXES — 1 MED (URL substring) + 2 LOWs
- **Codex 5.5:** NEEDS_FIXES — 2 MEDs (URL substring + `.env.template`) + 2 LOWs
- **Opus 4.6:** CLEAN with 3 LOWs
- **Opus 4.7:** CLEAN with 3 LOWs
- **Gemini 3.1 Pro:** CLEAN, recommends merge

## Merge recommendation

**Apply 3 tiny fixes (R5-M1 + R5-M2 + comment-#511) in R5 fix-round, then merge after a R6 sanity check.** Total wall-clock ~10-15 min.

Convergence is near-complete. After 5 rounds of cross-model review, the heartbeat refactor is well-tested, well-documented, with 5 reviewers explicitly recommending merge. The remaining concerns are minor and bounded.

claude-do:
# R6 Super-Swarm Synthesis — PR #503 (head `ec9fa58`)

**Models run:** Codex 5.3 ✓ | Codex 5.4 ✓ | Codex 5.5 ✓ | Opus 4.6 ✓ | Opus 4.7 ✓ | Gemini 3.1 Pro ✓
**All 6 reviewers VERDICT: CLEAN — MERGE-READY**

## Summary

**6/6 UNANIMOUS CLEAN.** After 6 rounds of cross-model review (codex × 3 + opus × 2 + gemini, repeated each round), the heartbeat refactor has converged. Every R5 fix VERIFIED FIXED with file:line evidence. Zero new HIGH or MEDIUM findings. 3 minor LOWs across all reviewers, all defer-OK.

## Findings (all LOW, defer-OK)

| Finding | Models | Action |
|---|---|---|
| Missing direct test for malformed URL parse edge cases (`convex.cloud`, `https://`) | codex 5.3, codex 5.4 | Optional belt-and-suspenders test. Cheap to add but not required. |
| Missing direct test for hostname rewrite preserving port/path/query | codex 5.3 | Cheap optional test. |
| Doc drift in scheduler comment + runner README: "restart required" for interval change is now technically stale since scheduler reads on-chain each loop | codex 5.4 | Restart still affects observability (`heartbeatIntervalSeconds` cached at boot for status reporting). Update wording in follow-up doc PR. |

## R5 fixes verification (6/6 VERIFIED FIXED)

| R5 finding | All reviewers agree |
|---|---|
| **R5-M1**: URL hostname-suffix check | VERIFIED FIXED at `runnerCastHeartbeat.ts:240-255` |
| **R5-M2**: `.env.template` commented out | VERIFIED FIXED at `.env.template:47-51` |
| **R5-LOW**: Comment references `#511` | VERIFIED FIXED at `heartbeatScheduler.ts:150-154` |

## Per-model verdicts

- **Codex 5.3:** CLEAN — 2 LOWs (test coverage)
- **Codex 5.4:** CLEAN — 1 LOW (doc drift on restart-required wording)
- **Codex 5.5:** CLEAN — no findings
- **Opus 4.6:** CLEAN — no findings, thorough verification table
- **Opus 4.7:** CLEAN — no findings
- **Gemini 3.1 Pro:** CLEAN — recommends MERGE

## Convergence history

- R1: 2 HIGHs + 4 MEDs → fix `2dd279b`
- Cloud: 1 HIGH + 4 MEDs → fix `ab3bd4f`
- R2: 4 NEW MEDs → fix `d3df280`
- R3: 2 NEW HIGHs + 3 MEDs → fix `bf26ea8`
- R4: 2 NEW MEDs + 1 LOW + DISPUTED HIGH → fix `5227f80`
- R5: 2 MEDs (cross-model split) + 1 LOW → fix `ec9fa58`
- **R6: CLEAN — MERGE READY**

## Final test counts

- 152 runner tests pass (grew from 130 baseline through 6 rounds of fixes; +22 new tests for all the fix-round behaviors)
- 82 server tests pass
- All typechecks green

## Merge recommendation

**APPROVE & MERGE TO `dev`.** Awaiting Liam's call.

PR #543, opened 2026-05-22
feat(elder-runtime): Node supervisor — command-bus poll + tmux dispatch (#352)

claude-do:
## Summary

The KEYSTONE of Bundle 2. Hybrid cherry-pick of PR #416's elder-runtime supervisor per exploration audit (`~/claudes-world/tmp/20260521-2247-pr416-redo-guidance.md`).

A Node.js supervisor process that runs in each elder container, polls the Convex command-bus (live as of PR #538) for new commands, and dispatches them via `tmux send-keys` to the Claude REPL inside the container.

## Changes

- `packages/elder-runtime/` — new workspace package (~1100 LOC TypeScript across 21 files)
- Supervisor binary entry point (`packages/elder-runtime/src/main.ts`)
- Convex client wrapper (`convexClient.ts`)
- tmux dispatch logic (`tmuxSink.ts`)
- Heartbeat upsert loop (`heartbeat.ts`)
- 6 command handlers (user_message, system_message, snapshot_request, reset, freeze, unfreeze)
- Graceful shutdown / signal handling + singleton PID-file lock
- 5 vitest test files, 17 tests, pure-unit (no container)
- `agents/Dockerfile` — wires supervisor binary into the image (`tsx` + `COPY packages/elder-runtime/` + `chown elder`)
- `agents/entrypoint.sh` — starts supervisor in bg + foreground monitor loop
- `apps/server/convex/commandBus.ts` — supervisor-required signature deltas: `claimNext` returns full doc (not just `_id`), `releaseLease` mutation added (for freeze-gate fall-through)

## Cherry-pick stack (oldest → newest)

1. `8c82df2a` — feat: Node supervisor — command-bus poll + tmux dispatch + health
2. `55f23ed` — fix: claimNext returns full doc + snapshot cap + heartbeat guard + config hardening
3. `f3bd9d6` — fix: freeze blocks user_message+system_message with failCommand reason=frozen
4. `22b8d20` — fix: tmux+ttyd boot + nonce protocol + workspace dep (super-swarm R1)
5. `4d11620` — fix: bracketed-paste nonce + releaseLease freeze + readiness file + timeout (super-swarm R2)
6. `e448037` — fix: healthcheck literals + writable readiness + nonce occurrence-count (super-swarm R2)
7. `23bce12` — fix: FAIL marker seeded in prompt + narrow singleton-lock catch (super-swarm R3)
8. `4e91698` — chore: regenerate pnpm-lock.yaml

[...]

PR #545, opened 2026-05-22
fix(elder-runtime): tmuxSink.loadBuffer — execFile.input silently dropped (production-breaking)

claude-do:
## Production-breaking bug in just-merged PR #543

**Caught by:** Bundle 3 exploration agent post-#543-merge (~04:25 ET 2026-05-22)
**Missed by:** All 4 swarm rounds on PR #543 including Tier 1 R2's 18-test empirical pass

### The bug

`packages/elder-runtime/src/tmuxSink.ts:18` was:

```typescript
await execFileAsync("tmux", ["load-buffer", "-b", name, "-"], { input: content } as any);
```

Per Node docs: `input` is documented on `execFileSync` (sync variant) but NOT on async `execFile`. The async version silently ignores it. The `as any` cast hid the type-level signal.

### Failure mode

`tmux load-buffer -b NAME -` reads buffer content from stdin. With no input piped, tmux receives EOF immediately, exits 0 with empty buffer. Then `pasteBuffer` pastes nothing into Claude's tmux pane. Supervisor's command delivery silently goes nowhere. Healthcheck (`pgrep -x claude`) reports healthy because Claude is alive — just unfed.

Bundle 2's keystone supervisor was shipping broken on dev-containerize-agents.

### Why swarm missed it

Unit tests mocked `execFile`, so they didn't exercise the runtime behavior. The 3-tier swarm caught semantic bugs (freeze-deadlock, lease/nonce timeout, Convex signature alignment) but the sync/async option-name overlap was opaque to all reviewers including Tier 1's 18-test empirical pass.

**Architectural lesson:** test-against-actual-system-boundaries is the remaining safety net for runtime-API-misuse bugs.

### Fix

Replaced `execFileAsync` with `spawn` + `child.stdin.end(content, "utf8")` which correctly pipes content through stdin until EOF.

Added test that uses real spawn semantics (NOT mocked execFile) to catch this class of regression.

### Test plan

- [x] Workspace typecheck PASS
- [x] elder-runtime test suite PASS with new regression test
- [ ] Local 3-tier swarm
- [ ] Merge to dev-containerize-agents

## Refs

- PR #543 (merged 04:00 ET 2026-05-22, head `46f0fe6`, merge commit `b4ea1a8`)
- Bundle 3 exploration report: `~/claudes-world/tmp/20260522-bundle3-exploration.md`

### DIFF
--- a/packages/elder-runtime/src/tmuxSink.ts
+++ b/packages/elder-runtime/src/tmuxSink.ts
@@ -1,4 +1,4 @@
-import { execFile } from "node:child_process";
+import { execFile, spawn } from "node:child_process";
 import { promisify } from "node:util";
 
 const execFileAsync = promisify(execFile);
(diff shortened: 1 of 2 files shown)


claude-do:
**Codex review** · round 1 · head `ff705bf`

Verdict: CLEAN

Verified: spawn+stdin.end is correct canonical pattern; double-reject after Promise settlement is harmless (ignored); stdin handlers registered before stdin.end so no ordering race; default spawn stdio is pipe so child.stdin is non-null; utf8 encoding correct for multibyte content; queueMicrotask test timing is sound (handlers registered synchronously before Promise yields).

#codex

PR #554, opened 2026-05-22
feat(caddy): dockerized Caddy router for app.clan-world.com (#348 v3)

claude-do:
## Summary

Third attempt at #348 — Caddy host-imported snippet for `app.clan-world.com` ttyd routing. **Replaces PR #546's host-Caddy-snippet design**, which the 3-tier swarm blocked on two CRITICAL findings (RCE risk from \`--writable\` ttyd + top-level site block never receives traffic via host caddy's \`:18080\` block).

Per Liam's 16:02 ET voice 16521 direction — dedicated dockerized Caddy alongside the host Caddy, so routing config lives in docker-compose and is portable for team members on their own Cloudflare setups.

## Architecture

- **NEW** \`caddy\` compose service (caddy:2-alpine) bound to \`127.0.0.1:\${CADDY_HOST_PORT:-18081}:80\`
- **NEW** \`agents/shared/caddy.conf\` — single \`:80\` site, \`handle_path /elder-N/*\` → \`elder-N:7681\` via Docker DNS
- **NEW** \`agents/shared/README-caddy.md\` operator doc (prod cloudflared setup + local dev + CLAN_WORLD_WEB_UPSTREAM)
- Cloudflared edit needed: one ingress entry routing \`app.clan-world.com\` → \`http://127.0.0.1:18081\` (runbook Step 8 walks through it)
- ttyd \`--writable\` REMOVED — closes C-1 RCE; operators send commands via command bus
- Cloudflare Access policy at perimeter (matches existing cockpit pattern)

## What's discarded vs PR #546

- Host caddy snippet install script
- Top-level \`app.clan-world.com { ... }\` block (root cause of C-2)
- Per-elder \`ELDER_N_TTYD_PORT\` host-loopback mappings
- Unauthenticated writable ttyd

## What's kept vs PR #546

- /elder-N/ public URL pattern
- \`handle_path\` stripping
- WebSocket-conscious proxying
- Frontend URL updates (TerminalTab + useConnectionStatus + Playwright)

## Test plan

- [x] \`docker compose --profile dev config --quiet\` (dummy env) — passed
- [x] \`caddy validate --config agents/shared/caddy.conf --adapter caddyfile\` — passed
- [x] \`git diff --check\` — passed
- [ ] Local 3-tier swarm review
[...]

PR #553, opened 2026-05-22
Bundle 3: dev-phase-3-final-polish — final polish (runbook + Makefile + commandBus + security)

claude-do:
## Summary

Stacked release PR (Bundle 3 → Bundle 2). 4 polish commits + 1 BLOCKED Caddy PR atop the Bundle 2 keystone work.

## What's in Bundle 3

| Sub-PR | Title | Status |
|---|---|---|
| #547 | Phase 2 migration runbook (#356) — 3 docs + rehearsal compose | ✅ MERGED |
| #548 | agents/Makefile + .docker-mounts scaffolding (#355) — operator entrypoint | ✅ MERGED |
| #549 | commandBus surviving #421 fixes — 5 correctness ports + 10 tests | ✅ MERGED |
| #551 | settings.json secret-exfil deny additions — defense-in-depth | ✅ MERGED |
| #546 | Caddy host-imported snippet (#348) | ⛔ BLOCKED architectural |

## #546 status (BLOCKED on architectural decision)

PR #546 has 2 CRITICAL findings from 3-tier swarm:

1. **C-1 (RCE risk):** \`agents/entrypoint.sh\` launches ttyd with \`--writable\` + no auth; snippet proxies to public \`app.clan-world.com/elder-N/\`. Need: Cloudflare Access zero-trust? Caddy basicauth? Revert ttyd to \`--readonly\` regardless?
2. **C-2 (won't route):** Snippet defines top-level \`app.clan-world.com { ... }\` block but host caddy wraps \`*.clan-world.com\` inside \`:18080 { bind 127.0.0.1 }\` because cloudflared terminates TLS at loopback. Verified against actual \`/etc/cloudflared/config.yml\` + \`ss -tlnp\`. Install + validate green but production traffic never reaches the block.

See \`docs/reviews/pr546-synthesis.md\` for full triage.

Bundle 3 can ship to Bundle 2 WITHOUT #546 — the other 4 polish PRs are independent. #546 can land as a separate follow-up once architectural shape is decided.

## Why stacked on Bundle 2 (not direct to dev)

Per Liam's directive 2026-05-22 voice (~15:32 ET): three logical groups of work as separate stacked release PRs so each is independently reviewable.

## Test plan

- [x] All sub-PRs swarm-reviewed individually
- [x] All sub-PR CI green
- [ ] Liam UAT
- [ ] Liam merge (Bundle 3 → Bundle 2)
- [ ] Liam decision on #546 architectural shape (separate from this PR's review)

## Refs

[...]

PR #552, opened 2026-05-22
Bundle 2: dev-containerize-agents — elder runtime + supervisor + agent infrastructure

claude-do:
## Summary

Stacked release PR (Bundle 2 → Bundle 1). 29 commits including the elder-runtime Node supervisor, command-bus + agent communication schema, heartbeat container with HMAC webhook, agent Docker image + shared layout, tmuxSink correctness hot-fix, and all per-PR swarm fix-rounds.

## What's in Bundle 2

Headline merges (each was itself a sub-PR through 1-4 swarm rounds):

- **PR #533 — agents/shared layout** — per-elder runtime workspace with home-claude settings + run.sh + bootstrapped tmux session structure
- **PR #534 — URL scheme rename (#354)** — cockpit → root, map → /map (frontend nav + e2e test sweep)
- **PR #542 — command-bus schema + mutations (#351)** — agentCommands + elderHeartbeat + commandResults tables, claimNext / ackCommand / completeCommand / failCommand mutations, sweepStaleDelivered cron
- **PR #542 — elder container image Dockerfile (#345)** — clanworld/agents:dev image, multi-stage build, healthcheck via pgrep
- **PR #543 — elder-runtime Node supervisor (#352)** — command-bus poll + tmux dispatch + lease/nonce/freeze handshake + R1+R+1 fix-rounds
- **PR #545 — tmuxSink execFile.input hot-fix** — critical production-breaking sync-vs-async API misuse caught post-#543-merge

Plus heartbeat container, secret materialization scaffolding, indexer flag wiring, anvil-fork dev-mode, .env.template alignment, multiple swarm-driven fix-rounds.

## Why stacked on Bundle 1 (not direct to dev)

Per Liam's directive 2026-05-22 voice (~15:32 ET): three logical groups of work as separate stacked release PRs so each is independently reviewable.

- Bundle 1 (#532): dev-containerize-services → dev — Convex self-hosting + heartbeat container + anvil-fork
- Bundle 2 (this PR): dev-containerize-agents → dev-containerize-services — agent containerization + command-bus + supervisor
- Bundle 3 (next PR): dev-phase-3-final-polish → dev-containerize-agents — final-polish docs/correctness/security

## Test plan

- [x] All sub-PRs swarm-reviewed individually before merge
- [x] PR #543 + #545 keystone supervisor + hot-fix landed clean
- [x] All CI green on integration branch
- [ ] Liam UAT
- [ ] Liam merge (per ADR 0018, integration→dev merges are Liam-only)

[...]

PR #551, opened 2026-05-22
fix(agents/settings): add 13 secret-exfil deny entries from PR #421 opus 4.7 addendum

claude-do:
## Summary

Bundle 3 final security polish. Port the secret-exfil deny entries from the original PR #421 opus 4.7 addendum (commit \`8630061\`, never merged — that recover branch was closed-stale during the 2026-05-16 phase-0 rollback).

## Why

Current HEAD's \`agents/shared/home-claude/settings.json\` has good \`Write\`/\`Edit\` denies on \`~/.claude/settings.json\` + \`CLAUDE.md\`, and existing \`cat //proc/*/environ\` + \`env\` + \`printenv\` denies. But it's missing:

1. **Docker secret mount denies** — the new self-hosted infrastructure mounts secrets at \`/run/secrets/bus-elder-N\`, \`/run/secrets/bus-operator\`, \`/run/secrets/webhook-shared\`, etc. A prompt-injected agent without these denies could read its peers' bus secrets and impersonate them.

2. **Inline-script and env-dump shell-verb denies** — \`Bash(strings *)\` can binary-grep \`/proc/*/environ\`; \`Bash(set)\`/\`Bash(declare *)\`/\`Bash(export)\` are shell built-ins that dump environment; \`Bash(node -e *)\`/\`Bash(python -c *)\`/\`Bash(python3 -c *)\` are inline scripts that can read \`process.env\`, the filesystem, or \`/run/secrets/\` directly.

Defense-in-depth: these complement the existing cat/env denies for credential exfiltration resistance.

## Diff scope

13 lines added to \`permissions.deny\` array. All other settings preserved (allow, ask, hooks, etc. untouched).

## Test plan

- [x] \`cat agents/shared/home-claude/settings.json | python3 -m json.tool > /dev/null\` — passed (valid JSON)
- [x] \`git diff\` — purely additive
- [ ] Local 3-tier swarm review
- [ ] Merge to \`dev-phase-3-final-polish\`

## Refs

- Original PR #421 (recover/release-superswarm-r1, closed-stale)
- Bundle 3 exploration: \`~/claudes-world/tmp/20260522-bundle3-exploration.md\`
- Closes 1 of 18 PR #421 findings; remaining: HMAC heartbeat (deferred for Liam's morning review)
- Sibling Bundle 3 PRs: #547 (runbook MERGED), #548 (Makefile MERGED), #549 (commandBus MERGED), #546 (Caddy BLOCKED architectural)

[...]

PR #560, opened 2026-05-23
Dev

mcorrig4:
@claude-do pls create changelog for release candidate and super swarm review


mcorrig4:
@codex please review

claude-do:
# Phase Super-Swarm Synthesis — PR #560 (head 09f78c8)

**Models run:** Codex 5.3 ✓ | Codex 5.4 ✓ | Codex 5.5 ✓ | Opus 4.6 ✓ | Opus 4.7 ✓ | Gemini 3.1 Pro ✓

**NOTE on Opus 4.6:** Marked "CLEAN" but with 5 MEDIUMs (no HIGHs). It did NOT detect the merge-order bug — it described Bundle 3 components (Makefile bootstrap, Caddy snippet) as if present. Per memory `feedback_super_swarm_cross_model_disagreement_resolution`: verify against actual code. The merge-order bug is verified real via `git log origin/dev..origin/dev-containerize-agents` (11 commits stranded). Opus 4.6's MED-level findings are still useful additions.
**Phase:** dev → main (v2.14.0 release combining Bundles 1+2+3)
**Diff size:** 7716+/389- across 103 files (9771 diff lines)

## Summary

**NEEDS_FIXES — BLOCKING merge-order bug.**

The super-swarm caught a critical issue: **PR #560's diff is incomplete**. Bundle 3's work (Makefile operator entrypoint, dockerized Caddy v3, command-bus runbook, Bundle 3 cloud-review fixes) is stranded on `origin/dev-containerize-agents` and was never merged into `origin/dev`. PR #560 (`dev → main`) therefore ships an incoherent state where the elder-supervisor + command-bus auth gates are present but the operator workflow + reverse-proxy + dockerized Caddy that make them executable are missing.

**Three codex reviewers (5.3, 5.4, 5.5) all independently surfaced symptoms of this same root cause:**

- Docs reference `make up`, `make status`, `make reset-elder-N`, `make bootstrap-bus-secrets` — none of these targets exist in dev's `Makefile`. They exist in `agents/Makefile` on `dev-containerize-agents`.
- Compose declares per-elder `bus-elder-N` and `bus-operator` secret mounts — but the bootstrap path to generate those secret files and push the matching env vars into the self-hosted Convex deployment is absent from dev.
- The release brief claims dockerized Caddy v3 routing — no Caddy service exists in `docker-compose.yml` on dev. Caddy file `agents/shared/caddy.conf` exists only on `dev-containerize-agents`.

## Root cause (verified via git)

The merge timeline:
- **15:23:05 UTC** — PR #532 (`dev-containerize-services` → `dev`) merges Bundle 1
- **15:23:29 UTC** — PR #552 (`dev-containerize-agents` → `dev`) merges Bundle 2
- **15:23:45 UTC** — PR #553 (`dev-phase-3-final-polish` → `dev-containerize-agents`) merges Bundle 3 — but PR #552 has **already** merged a few seconds earlier, so Bundle 3 lands in `dev-containerize-agents` after that branch had already been integrated to dev.

Net effect: 11 commits are stranded on `dev-containerize-agents` beyond the SHA that became dev. Verified via `git log --oneline origin/dev..origin/dev-containerize-agents`.

The stranded commits include:
- `2de51e1` (Bundle 3 release merge)
- `bf9b809` + `34e8107` + `f21112f` + `e85a732` + `72bc366` (PR #554 dockerized Caddy v3 + cloud-review + R1/R2 fix-rounds)
- `ddbb7fe` + `26035e8` (PR #557 Bundle 3 cloud-review fixes)
- `275cc93` (PR #551 settings deny entries)
- `9eb16f3` (PR #549 command-bus survivors)
- `beb17c3` (PR #548 Makefile scaffolding)
- `5eee668` (PR #547 migration runbook)

26 files / +2061 LOC / -291 LOC total stranded.

## Recommended fix path (per ADR 0018)

1. **Open a new release PR** `dev-containerize-agents → dev` for the 11 stranded commits.
2. **Liam merges that PR** (Liam-only on dev per ADR 0018).
3. **Re-cycle PR #560** so the dev→main diff now includes the complete Bundle 1+2+3.
4. **Re-run super-swarm** on the updated PR #560 to verify no NEW integration seam bugs surface after the merge.

Alternative paths (NOT recommended):
- ❌ Cherry-pick Bundle 3 commits onto dev directly — breaks gitflow + loses merge audit trail.
- ❌ Close #560 + open new PR `dev-containerize-agents → main` — skips dev as the integration branch (violates ADR 0018).

## Codex synthesis (root-cause symptoms)

Once the merge-order is fixed, these specific findings need verification against the updated diff:

### Symptoms tied to "missing Bundle 3" (resolve when bundle 3 lands in dev)

| Severity | Models | Surface | Symptom |
|---|---|---|---|
| HIGH | 5.4, 5.5 | `docker-compose.yml` (compose) | No `caddy` service, no `agents/shared/caddy.conf`. Cockpit + per-elder ttyd routes unreachable. |
| HIGH | 5.4, 5.5 | `agents/README.md`, `Makefile`, `.env.template` | Docs reference `make up`/`status`/`logs`/`pause-elder-N`/`reset-elder-N`/`wipe-elder-N`/`bootstrap-bus-secrets`/`oauth-bootstrap` — none exist in dev. (All present in dev-containerize-agents' `agents/Makefile`.) |
| HIGH | 5.4, 5.5 | `apps/server/convex/commandBus.ts` (auth gate) | `BUS_OPERATOR_SECRET` + `BUS_ELDER_SECRET_N` env vars are validated against `process.env` on every call but there's no bootstrap step setting them on the self-hosted Convex backend. Bundle 3's `bootstrap-bus-secrets` Makefile target generates secret files + pushes them via `convex env set` — currently absent from dev. |

### Opus 4.7 NEW HIGHs (apply independent of merge-order fix — present in BOTH dev and dev-containerize-agents)

| Severity | Models | Surface | Finding |
|---|---|---|---|
| HIGH | 4.7 | `agents/entrypoint.sh:43` + `agents/init-firewall.sh:50-55` + `docker-compose.yml:30-31` | **ttyd `--writable` + bridge-network ACCEPT = cross-elder paste vulnerability.** ttyd runs writable with no auth; firewall permits INPUT from the entire bridge subnet; all four elders share `clan-world-internal`. So `elder-2` can `curl elder-1:7681/ws/...` and inject keystrokes into elder-1's claude pane. The bracketed-paste nonce protocol doesn't help — anyone on the bridge can speak ttyd directly. Caddy is north-south only; doesn't isolate east-west. Fix: `ttyd --interface 127.0.0.1` (docker-exec only) OR `ttyd --credential <user>:<bcrypt>` from a secret, OR iptables OUTPUT rules blocking peer ttyd ports. |
| HIGH | 4.7 | `packages/elder-runtime/src/freezeGate.ts` + `agents/entrypoint.sh:82-84` + `docker-compose.yml:36` | **Freeze gate is in-memory, silently lost on supervisor restart.** `restart: unless-stopped` brings the elder back unfrozen on any crash / OOM / `docker restart` — no operator signal that the kill-switch was bypassed. Fix: persist gate to `${stateDir}/freeze.flag` on every transition; read it on supervisor boot before starting the poll loop. |
| HIGH | 4.7 | `packages/elder-runtime/src/main.ts:132` + `packages/sdk/convex/schema.ts:8881` + `agents/heartbeat/README.md:33` | **`lastTickProcessed` increments per command, not per game tick.** Counter advances on every dispatched command (user_message, freeze, snapshot_request, reset, etc.), but schema field name + runbook framing imply game-tick progress. Observability dashboards alarming on tick stalls will be falsely green (operator sent pings) or falsely red (no ops events but elder doing tick work). Cross-PR seam bug — Bundle 2 named the field, Bundle 3's schema took it verbatim. Fix: rename to `commandsProcessed` everywhere OR wire `lastTickProcessed` to actual chain tick progress. |

### Opus 4.7 NEW MEDIUMs (apply post-merge-order-fix)

| Severity | Models | Surface | Finding |
|---|---|---|---|
| MED | 4.7 | `agents/elder:14-19` + `agents/shared/APPENDED_SYSTEM_PROMPT.md:30-37` | **`agents/elder` is a stub that prints "not implemented" and exits 0.** APPENDED_SYSTEM_PROMPT tells the Elder its only game interface is `elder world snapshot`, `elder clan view`, etc. Every Elder will boot, try `elder world snapshot`, see the stub, and have no way to act. Fix: stub should `exit 1`; system prompt must disclose stub state. |
| MED | 4.7 | `packages/elder-runtime/src/commandHandlers/reset.ts:11-17` + `agents/shared/run.sh:122-125` | **`reset` doesn't reset — it respawns the same conversation.** Reset calls `tmux.respawnPane()` which re-execs `run.sh`; run.sh always `claude --continue` when prior session JSONLs exist; `runtime/elder-N/.claude/` is a persistent volume. Operator expecting clear loses confidence. Fix: rename to `restart` OR implement real clear (remove session JSONLs). |
| MED | 4.7 | `apps/server/convex/commandBus.ts` (all mutations) | **Convex `secret` args are visible in dashboard log payload.** Every bus call passes `secret: v.string()` as arg; Convex dashboard logs args object on each invocation. Any operator with dashboard auth can read every bus secret out of historical logs. Fix: pass via `ctx.request.headers` OR ConvexAuth identity. |
| MED | 4.7 | `agents/init-firewall.sh:73-87` | **Anthropic IPs resolved once at container start, no refresh.** `getent ahostsv4 api.anthropic.com` → pinned via iptables. Anthropic uses CDN; IPs rotate. When a cached IP drops from rotation, elder loses Claude API connectivity until `docker restart`. For multi-day live demos = silent multi-hour outage. Fix: cron inside container re-running allow-host stanza every ~10 min, OR `ipset` tracking live A-records. |
| MED | 4.7 | `.env.template:277-280` + `agents/elder-1/.env.template:23-26` + `agents/shared/run.sh:34-44` | **Two parallel mechanisms for `CLAUDE_CODE_OAUTH_TOKEN` with undocumented precedence.** Per-elder env_file vs top-level overrides. If both set with different values, behavior depends on docker-compose merge order — easy to leak/misroute tokens. Pick one source of truth. |
| MED | 4.7 | `packages/elder-runtime/src/tmuxSink.ts:35-39` + `userMessage.ts:53-55` | **Bracketed-paste assumption on claude-code TUI version.** `\e[200~...\e[201~` framing only works if claude-code TUI is in bracketed-paste-aware mode. Respawn during a non-TUI prompt = framing bytes appear as literal text + NONCE marker breaks line-anchored regex. Fix: pin claude-code version + add integration test asserting marker matches against real output. |
| MED | 4.7 | `apps/server/convex/commandBus.ts:23-67` | **`enqueueCommand` has no rate-limiting.** Leaked operator secret = unbounded bus floods exhausting elder claude budget. Pair with M4 (secrets in dashboard logs) → real blast radius. Fix: per-source-window cap (10/min/source) + max payload size. |

### Gemini 3.1 Pro NEW HIGHs (apply independent of merge-order fix — present in dev AND dev-containerize-agents — VERIFIED against actual code)

| Severity | Models | Surface | Finding |
|---|---|---|---|
| HIGH | gemini-3-1-pro | `packages/elder-runtime/src/tmuxSink.ts:35` + `commandHandlers/userMessage.ts:30-36` | **Bracketed-paste prompt injection vulnerability.** `payload.text` (untrusted operator content) is concatenated with the nonce instruction + passed to `loadBuffer` + `pasteBuffer({ bracketed: true })`. If `text` contains the bytes `\x1b[201~` (end-bracketed-paste sequence), the paste terminates early + remaining content executes as raw keystrokes in the claude-code tty. **Verified real:** zero sanitization or escape applied to `payload.text` before paste. Fix: strip/escape `\x1b[201~` (and adjacent escape sequences) from text before loadBuffer; OR pass `payload.text` to claude via stdin to a `claude --print` invocation rather than tmux paste; OR add an integration test asserting paste-injection cannot break out of bracketed mode. |
| HIGH | gemini-3-1-pro | `packages/elder-runtime/src/main.ts:28-64` | **TOCTOU race in supervisor.lock between openSync and writeSync.** Process A: `openSync(lockPath, "wx")` succeeds, hasn't written PID yet. Process B: `openSync` throws EEXIST, reads empty lockfile, `parseInt("", 10) = NaN`, `Number.isFinite(NaN) = false`, skips pid-check block, falls through to `unlinkSync(lockPath)` + `openSync(lockPath, "wx")` + writes its own PID. Both processes now think they hold the lock. **Verified real:** the `if (Number.isFinite(stalePid))` branch on line 35 is the only gate before the unlink. Practical risk narrow (supervisor.lock acquired once per container boot, normal compose flow is single-supervisor-per-container) but the bug is on the singleton-lock-correctness path the previous super-swarm rounds invested in hardening. Fix: write-then-rename (atomic), OR add a lockfile-content-validity check (must contain a valid integer) before treating as stale, OR use `flock(2)` instead of file-based PID locking. |

### Gemini 3.1 Pro NEW MEDIUMs (post-merge-order-fix)

| Severity | Models | Surface | Finding |
|---|---|---|---|
| MED | gemini-3-1-pro | `docs/plans/dockerize-elder-infra-v1.md:5246` | Phase 2 cutover brings up compose (incl `heartbeat` container) while legacy systemd units still running for ~30-min window. Violates "ONLY one active heartbeat caller" invariant. Fix: documented sequencing — stop systemd-heartbeat BEFORE compose-heartbeat starts, OR add an explicit lockfile both honor. |
| MED | gemini-3-1-pro | `docker-compose.yml` (Caddy/dashboard auth — only relevant once Bundle 3 merges in) | Caddy basicauth doesn't natively support `_FILE` env var pattern. Mounting `/run/secrets/dashboard-basicauth` as `CONVEX_DASHBOARD_BASIC_AUTH_HASH_FILE` will leave dashboard unauthenticated or broken. Fix: render the auth hash into Caddyfile at boot via templating OR switch to a different auth scheme. (Applies post merge-order-fix.) |
| LOW | gemini-3-1-pro | `apps/server/convex/crons.ts:45` + `apps/server/convex/commandBus.ts:223-253` | Cron + function named `sweepStaleDelivered` but no `delivered` status exists in schema — function actually sweeps `leased`/`acked` expired-lease rows. Naming legacy from earlier design. Either rename function to `sweepExpiredLeases` (or similar) OR document the legacy name. |

### Opus 4.6 NEW MEDIUMs (apply post-merge-order-fix)

| Severity | Models | Surface | Finding |
|---|---|---|---|
| MED | 4.6 | `apps/server/convex/commandBus.ts:8,17` + `apps/server/convex/heartbeat.ts:127` | **Non-constant-time auth comparison** (pre-existing debt, flagged in PR #532 + unfixed). `checkOperatorAuth` and `checkElderAuth` use `!==` for secret comparison; webhook auth same. Timing side-channel leaks prefix/length. Low practical risk with high-entropy secrets, but pattern shipping to main for third PR in a row. Fix: `crypto.timingSafeEqual` after length-normalizing buffers, or HMAC compare. |
| MED | 4.6, 4.7 (overlap) | `apps/server/convex/commandBus.ts:32` | **`enqueueCommand` payload is `v.any()` with no size guard.** Operator-secret holder (or compromised orchestrator) can insert multi-MB payloads bloating `agentCommands` table. Fix: structured union OR explicit `JSON.stringify(payload).length` byte-length check. (Opus 4.7's M8 covers the rate-limiting angle of the same surface.) |
| MED | 4.6 | `agents/shared/home-claude/settings.json:15-21` | **CC permission deny-list has env-var exfiltration gaps.** Deny rules block `env`, `printenv`, `cat /proc/*/environ` but DON'T cover `set`, `declare -p`, `echo $VAR`, `compgen -v`. If CC's default in headless mode auto-allows unmatched Bash commands, the deny-list alone is insufficient to prevent `BUS_ELDER_SECRET_*` exfiltration via shell variable expansion. Fix: verify CC headless permission default OR add explicit `Bash(echo *)`, `Bash(set)`, `Bash(declare *)`, `Bash(compgen *)` deny rules. |
| MED | 4.6 | `packages/elder-runtime/src/convexClient.ts` | **ConvexHttpClient has no retry/backoff.** 15s timeout wrapper but zero retry. Transient Convex outage = current command fails immediately + `consecutiveErrors` ticks. Lease-expiry sweep (60s cron) eventually re-queues but elder loses 5+ minutes. Fix: 1-2 retries with jitter for transient HTTP errors (5xx, ECONNREFUSED, timeout). |

### Opus 4.6 NEW LOWs

| Severity | Models | Surface | Finding |
|---|---|---|---|
| LOW | 4.6 | `docs/plans/dockerize-v1-revision-notes.md` vs `commandBus.ts:117` | Plan doc says `acked` status REMOVED, implementation retains it (correctly — load-bearing for freeze-before-ack gate + lease-expiry sweep). Doc is stale. Update revision notes. |
| LOW | 4.6 | `packages/elder-runtime/src/commandHandlers/userMessage.ts:36` | `pasteBuffer` target is `config.elderId` while `sendKeys` uses `this.session`. Asymmetry is fragile (today equal via `SESSION_NAME=${ELDER_ID}`). Normalize to single source. |
| LOW | 4.6 | `apps/server/convex/commandBus.ts:74` | `broadcastSequence` undefined field on non-broadcast inserts pollutes `by_broadcast_sequence` index. Omit field entirely on non-broadcast. |

### Codex MEDIUMs + LOWs (apply post-merge-order-fix)

| Severity | Models | Surface | Finding |
|---|---|---|---|
| MED | 5.3, 5.4, 5.5 | `agents/shared/run.sh:47`, `packages/elder-runtime/src/config.ts:25`, elder-N templates | Stale `BUS_ELDER_SECRET` env-var path. Docs + warning still tell operators to set this env var, but the runtime now reads `BUS_ELDER_SECRET_FILE` from the Docker secret mount. Warning is false-positive misleading. Fix: remove the env-var path from templates + run.sh warning. |
| MED | 5.3 | `apps/server/convex/commandBus.ts:44`, `packages/sdk/convex/schema.ts:446` | Broadcast sequence relies on optional-key ordering of the `by_broadcast_sequence` index. If undefined-key rows compete in ordering, batches can reuse sequence ids. Suggested fix: dedicated singleton counter table. |
| MED | 5.4 | `agents/shared/run.sh:22-25` | Elder session continuity uses path-encoding heuristic + `claude --continue`. Documents that path-encoding drift causes silent fallback to fresh session. Fix: persist explicit session-id + use `claude --resume <id>`. |
| MED | 5.5 | `docs/runbooks/self-hosted-convex.md:49` | Post-deploy env block uses bare `convex env set`, bypassing pinned `CONVEX_CLI_PINNED_VERSION` enforced by `bin/deploy-convex.sh`. Fix: use `npx -y convex@${CONVEX_CLI_PINNED_VERSION}` wrapper. |
| LOW | 5.3 | `apps/server/convex/commandBus.ts:210` | Numbering comments skip `7` → `8` → `9` → `10`. Renumber. |
| LOW | 5.5 | `packages/elder-runtime/src/main.ts:75` | Readiness file written before bus/tmux operations succeed. Move marker to after first successful heartbeat or `tmux has-session`. |
| LOW | 4.7 | `packages/elder-runtime/src/config.ts:8-14` | `parsePositiveInt` accepts trailing junk (`parseInt("5min",10)` → `5`). Use regex or `Number.isInteger`. |
| LOW | 4.7 | `packages/elder-runtime/src/main.ts:148` | Singleton lock cleanup not signal-driven. Add `process.on("exit", cleanupLock)`. |
| LOW | 4.7 | `packages/elder-runtime/src/commandHandlers/{userMessage,systemMessage}.ts` | Near-duplicate handlers. Extract shared `dispatchWithNonce` helper. |
| LOW | 4.7 | `agents/heartbeat/Dockerfile` | `COPY . .` builds whole monorepo. ~2GB image. File for cleanup post-hackathon. |
| LOW | 4.7 | `docker-compose.yml` elder healthchecks | Loose `pgrep -f 'tsx.*main.ts'` pattern; tighten to anchored regex. |
| LOW | 4.7 | `apps/web/vercel.json:3-6` | `permanent: true` on `/cockpit` → `/`. Once shipped, browser/CDN caches the 301. Consider 302 until bedded in. |
| LOW | 4.7 | `entrypoint.sh:67-69` | Warns + continues when tsx/elder-runtime missing. Half-functional elder passes docker health (claude alive) but never acts on commands. Fail-closed. |

## Cross-cutting observations

- The biggest phase-level issue is not inside an individual bundle; it is the seam between them — and the merge ordering meant that seam isn't actually wired up in the dev branch.
- Test coverage is unit-level at the edges. No integration test exercises the load-bearing chain `enqueueCommand → claimNext → ack/fail/complete → tmux dispatch → nonce detection`.
- Codex CLI sandbox is read-only → all three codex reviews were emitted to stdout, NOT written directly to `docs/reviews/`. Captured manually from logs.
- Opus 4.6 + 4.7 + Gemini 3.1 Pro reviews pending (background jobs still in flight at synthesis time).

## Recommended next action

**STOP — do not merge PR #560 as-is.** Liam (or orchestrator on Liam's behalf) opens `dev-containerize-agents → dev` release PR for the 11 stranded commits. Once merged, recycle PR #560 + re-run super-swarm.

Once dev contains Bundle 1+2+3:
- Re-verify codex's HIGH-2 (`make` targets) is resolved
- Re-verify codex's HIGH-3 (Convex env passthrough for bus secrets) by tracing `agents/Makefile bootstrap-bus-secrets` execution
- Apply codex's stale `BUS_ELDER_SECRET` env-var path cleanup (M-1 above) as a follow-up fix-round
- Defer the broadcast-sequence + session-id + Convex CLI pinning findings to backlog issues

---

_Synthesis written 2026-05-23 by orchestrator. Opus + Gemini reviews pending — will be re-folded into this doc when complete._

PR #612, opened 2026-05-25
fix(entrypoint): ttyd-refresh loop — fix stale display on cockpit reconnect

claude-do:
## Summary
- Adds a 2s `tmux refresh-client -S` loop in the elder container entrypoint
- Fixes the cockpit "stuck at reset theming" display bug Liam reported 2026-05-25
- No claude TUI change, no docker rebuild penalty

## Root cause
Claude TUI uses xterm **alternate screen**. ttyd 1.7.7 replays its captured byte buffer to every new client; the buffer contains only the rename+color slash commands from session start (which happened in the *normal* screen before claude entered alt-screen). On every browser reconnect, the cockpit shows the stale theming instead of the live TUI frame, until the next TUI delta arrives.

## Fix
`agents/entrypoint.sh`: background `while true; do tmux refresh-client -t <session> -S; sleep 2; done` forces tmux to push a fresh frame to all clients every 2s. `-S` also refreshes the status line. Watchdog warns (does not exit) if the loop dies — display staleness is a UX issue not a process failure.

Full diagnosis: `research/ttyd-reset-display-bug-2026-05-25.md` (investigation by general-purpose subagent 2026-05-25, 95k tokens).

## Test plan
- [x] Diff reviewed: only 16 lines added, no existing behavior changed
- [ ] Validate in dev: rebuild + restart one elder container, then disconnect+reconnect cockpit; expect to see live TUI frame within 2s
- [ ] Validate on prod: docker compose up -d after merge

### DIFF
--- a/agents/entrypoint.sh
+++ b/agents/entrypoint.sh
@@ -76,13 +76,23 @@ ttyd --port "${TTYD_PORT}" tmux attach-session -t "${SESSION_NAME}" &
 TTYD_PID=$!
 echo "[entrypoint] ttyd started on port ${TTYD_PORT} (PID ${TTYD_PID})"
 
-# 3. ttyd-refresh loop. Background tick that forces tmux to redraw the
-# alt-screen every 2s so new ttyd clients (or browser reconnects) see the
-# live frame instead of the stale replay buffer that contains only the
-# rename+color slash commands from session start.
+# 3. ttyd-refresh loop. Background tick that forces tmux to fully redraw
+# every attached client every 2s, so a browser that (re)connects sees the
+# live alt-screen frame instead of ttyd's stale replay buffer (which only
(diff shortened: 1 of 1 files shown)


claude-do:
**Codex** · round 1 · head `3ab40b4`

Verdict: NEEDS_FIXES (0 HIGH, 1 MEDIUM, 0 LOW)

- [MED] agents/entrypoint.sh:40 — `tmux refresh-client -S` only refreshes the status line (per manpage: "only update the client's status line"), not pane/alt-screen contents — the stale ttyd display bug is not fixed and the failure is silently swallowed by `|| true`

#codex

claude-do:
**Codex** · round 2 · head `239c611`

Verdict: CLEAN

Round-1 MEDIUM (refresh-client -S): RESOLVED

#codex

PR #622, opened 2026-05-26
fix(runner): bulletproof memory-wipe + branding reset reliability

claude-do:
## Problem

Two reset-reliability bugs surfaced during live elder research (Liam-reported 2026-05-26), root-caused + co-designed with codex (gpt-5.5, stress-test mode).

### Bug A — memory wipe silently skipped → uncontrolled auto-compact
The Convex subscription (`watchAuxiliary` / `onUpdate`) delivers the **latest snapshot**, not an event stream. While the runner spends seconds delivering one tick's prompt (paste + wait-for-hook-receipt), the tick clock advances and `onUpdate` fires once with the latest value — so the runner routinely jumps e.g. `2049 → 2052`, **stepping over** the scheduled wipe tick `2050`. `handleAuxiliaryUpdate` only checked the **exact current tick** (`tick % interval === 0`), so those wipes never fired. The session then ran unbounded until Claude Code **auto-compacted** (uncontrolled amnesia that defeats the deliberate wipe + continuity-file design). Observed: elder session jsonls bloating to 1.4–1.7 MB.

`startup.decideRestart` already guards this with the gap-aware `containsMemoryWipeTick`; the live loop did not.

### Bug B — branding `/color` left stuck unsubmitted
`brandElder` sends `/rename` then `/color` back-to-back. The fixed `delay(250)` + double-Enter raced the TUI: `/rename` submitted but `/color` was consistently left sitting in the input box (3/4 live elders). elder-2 also used the **invalid** color `olive`.

## Fix

- **`tickHandler.handleAuxiliaryUpdate`** is now gap-aware: it fires the reset when a wipe tick fell in `(lastDeliveredTick, currentTick]` (using the **confirmed-delivered** cursor threaded from `main.ts`), reason `memory_wipe_gap`, mirroring startup. Resets at the latest tick (snapshot-driven continuity, no replay). `wipeMarker` already protects mid-reset crashes.
- **`tmuxSink.sendSlashCommand`** now uses the same readiness + submit verification as tick pastes: `Escape` (clear stale slash/autocomplete state) → `prePasteReady` → type → **single** Enter → `postPasteSubmitted`, retry ≤3, **log-don't-throw** so a stuck command never strands the reset before the post-wipe continuity prompt.
- **elder-2** color `olive` → `cyan` (valid enum: `red|blue|green|yellow|purple|orange|pink|cyan|default`).

## Validation

- Hot-patched the live containers and ran the **actual patched `sendSlashCommand`** against the running elder sessions: elders 1–3 branded cleanly — `/color` now lands where it was previously stuck. (elder-4 hit a self-induced race from running the validation script *concurrently* with live tick delivery — a harness artifact; production branding runs synchronously inside `runResetFlow` with no concurrent delivery.)
- Added gap-aware wipe tests (exact tick, gap-crossing, no-cross) to `tickHandler.test.ts`.
- **All 60 runner tests pass; runner typecheck clean.**

## Notes
- Auto-compact intentionally **not** disabled (codex call): make scheduled wipe reliable first — auto-compact is the safety net. Follow-up: observe jsonl size at 30/40/50 ticks; shorten interval 50→40 if it still bloats.
- Live containers run the old runner module until restart/image rebuild; this PR is the durable path.

### DIFF
--- a/agents/shared/runner/elder-config.json
+++ b/agents/shared/runner/elder-config.json
@@ -1,6 +1,6 @@
 {
   "elder-1": { "displayName": "Storm Riders", "color": "blue", "glyph": "⚡" },
-  "elder-2": { "displayName": "Iron Guard", "color": "olive", "glyph": "⛨" },
+  "elder-2": { "displayName": "Iron Guard", "color": "cyan", "glyph": "⛨" },
   "elder-3": { "displayName": "Crimson", "color": "red", "glyph": "✦" },
   "elder-4": { "displayName": "Verdant Wardens", "color": "green", "glyph": "❦" }
 }
--- a/packages/runner/src/main.ts
+++ b/packages/runner/src/main.ts
@@ -30,6 +30,11 @@ async function main(): Promise<void> {
         ? (started.startupState.lastReceivedTick ?? -1)
         : started.startupState.tickClock.tick)
       : (started.startupState.lastReceivedTick ?? started.startupState.tickClock.tick - 1);
+    // Separate cursor for wipe handling: advances whenever a reset FIRES
+    // (confirmed or not), so an unconfirmed reset (the session is already
+    // killed + relaunched fresh) does not re-fire the wipe on every wake and
+    // livelock. Seeded from the post-startup baseline. (codex R1 HIGH)
+    let lastWipeHandledTick = lastTickDelivered;
     for await (const aux of convex.watchAuxiliary(ac.signal)) {
       if (ac.signal.aborted) break;
       const deps = {
@@ -43,7 +48,14 @@ async function main(): Promise<void> {
         if (aux.pendingMessages.length > 0) await handlePendingMessages(deps, aux);
         continue;
       }
-      const result = await handleAuxiliaryUpdate(deps, aux, lastTickDelivered);
+      const result = await handleAuxiliaryUpdate(deps, aux, lastWipeHandledTick);
+      // Advance the wipe cursor whenever a reset fired, confirmed or not: the
+      // session was already killed + relaunched, so re-firing the wipe on the
+      // next wake would livelock. A failed continuity-prompt delivery recovers
+      // via the next normal tick (codex R1 HIGH).
+      if (result.resetFired) {
+        lastWipeHandledTick = aux.tickClock.tick;
+      }
       // Only advance the "delivered" cursor on confirmed receipt — if the
       // hook didn't write tickReceiveLog within the resend cap, the next
       // tick subscription wake-up will see this tick still un-delivered and
--- a/packages/runner/src/main.ts
+++ b/packages/runner/src/main.ts
@@ -43,7 +43,7 @@ async function main(): Promise<void> {
         if (aux.pendingMessages.length > 0) await handlePendingMessages(deps, aux);
         continue;
       }
-      const result = await handleAuxiliaryUpdate(deps, aux);
+      const result = await handleAuxiliaryUpdate(deps, aux, lastTickDelivered);
       // Only advance the "delivered" cursor on confirmed receipt — if the
       // hook didn't write tickReceiveLog within the resend cap, the next
       // tick subscription wake-up will see this tick still un-delivered and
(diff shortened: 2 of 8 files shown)


claude-do:
**Codex review** · round 1 · head `bb9f8e2a`

Verdict: NEEDS_FIXES (1 HIGH, 1 MEDIUM, 0 LOW)

- [HIGH] packages/runner/src/tickHandler.ts:53 — Gap-triggered reset repeats indefinitely: `containsMemoryWipeTick(lastDeliveredTick, currentTick, interval)` remains true after `runResetFlow` returns `confirmed: false` since `lastTickDelivered` only advances on confirmed delivery. A tick jump like 49→52 will re-trigger reset on every subscription wake until a later tick is confirmed. Needs a "wipe handled through tick N" cursor or reset confirmation must advance the cursor.
- [MED]  packages/runner/src/tmuxSink.ts:103 — Escape sent before readiness check on every retry. If Claude is mid-generation, the Escape interrupts active output rather than just clearing stale input. `prePasteReady` runs after Escape so it cannot gate this. Readiness should be checked before sending Escape, or Escape gated on visible input/autocomplete state.

#codex

claude-do:
**Claude subagent** · round 1 · head `bb9f8e2a`
Verdict: CLEAN

Traced every flagged concern with full repo context; all resolve correctly:
- **Off-by-one**: `containsMemoryWipeTick` is `(last, cur]`; cursor advances only on `result.confirmed`, so the boundary wipe tick is excluded next iteration → no double-reset. First-live-tick seed (main.ts:28-32) sound in all four branches.
- **Leading Escape after fresh claude launch**: safe — `prePasteReady` (10s poll) is the built-in readiness wait; Escape-first also prevents `/color`-command accumulation across retries.
- **Log-don't-throw**: cosmetic-only; a genuinely stuck input still surfaces via the continuity prompt's `prePasteReady` failure → reset not confirmed → retried. Does not silently strand a reset.
- **Circular import**: none at runtime — `pasteVerification.ts` imports `TmuxSink` as `import type` (erased). `tsc --noEmit` clean.

3 LOW (non-blocking, full notes in tier-1 file):
- [LOW] tmuxSink.ts:113 — branding failure warns to stdout only, not `recordRunnerEvent` telemetry (cosmetic, but a chronic `/color` TUI break would be invisible).
- [LOW] test/tmuxSink.test.ts — net-new `sendSlashCommand` retry/log logic has no direct unit test (only `loadBuffer` covered; live-validated though).
- [LOW] test/tickHandler.test.ts:19 — no assertion that exact-tick path uses `reason: "scheduled"`.

#claude-subagent

claude-do:
**Gemini flash review** · round 1 · head `bb9f8e2a`

Verdict: CLEAN

No correctness bugs, type errors, null-check gaps, or logic bugs found. Three low-severity observations noted but none are blocking:

- [LOW] packages/runner/src/tickHandler.ts:40 — parameter named `lastDeliveredTick` in signature vs `lastTickDelivered` at call site — naming inconsistency, no functional impact
- [LOW] packages/runner/src/tmuxSink.ts:98 — 3-attempt retry loop with sleep(500) per step can add up to ~6s latency on a stuck command — deliberate trade-off per design
- [LOW] packages/runner/src/tmuxSink.ts:112 — log-don't-throw on 3-attempt failure means a stuck /rename or /color silently proceeds — also deliberate per "never strand the reset" design doc

#gemini-flash

claude-do:
**Gemini flash review** · round 2 · head `4e4d4694`

Verdict: CLEAN

#gemini-flash

claude-do:
**Claude subagent** · round 2 · head `4e4d4694`

Verdict: NEEDS_FIXES (1 H)

H — exact-tick wipe can still livelock on an unconfirmed reset. The R1 cursor only guards `crossedScheduled`; `exactScheduled` (`tick % interval === 0`) is computed independently, so after an unconfirmed reset at boundary `T` (cursor advanced to `T`, `lastTickDelivered` lags) any intra-tick reactive re-yield (pendingMessages/banditView/chainEvents at tick `T`, per `getRunnerAuxiliary`) re-enters and `exactScheduled` re-fires the wipe — repeats until the clock leaves `T`. Same livelock class R1 flagged; only the gap branch was closed, and the exact boundary is the more common scheduled-wipe path. Fix: gate firing solely on `containsMemoryWipeTick(lastWipeHandledTick, currentTick, interval)`; keep `exactScheduled` only for the reason label. Add a regression test: `handleAuxiliaryUpdate(deps, makeAux(50), 50)` must NOT reset (current "livelock guard" test uses tick 52 / cursor 52, 52%50≠0, so it never exercises a handled exact boundary). MED Escape fix verified correct.

#claude-subagent

Issue #675, opened 2026-06-14
UI: smaller ttyd terminal font + human-readable clan ETA / action name


claude-do:
Two web-UI fixes requested by Liam (voice 2026-06-14):

1. **TTYD terminal font too large.** The 4 elder terminals in the cockpit show only a handful of lines because ttyd renders at the xterm.js default (15px). Reduce the font so more text is visible.

2. **Clan tab ETA shows a raw number.** When a clansman is performing an action, the clan tab shows "ETA" followed by a raw tick/timestamp number. It should show the action name (e.g. "Chop Wood") + a human-readable ETA (formatted local clock time, e.g. "ETA 3:42:10 AM").

Tracking issue for branch feat/ui-ttyd-font-and-clan-eta.

PR #676, opened 2026-06-14
feat(web): smaller ttyd terminal font + human-readable clan ETA + action name

claude-do:
## Summary

Two ClanWorld cockpit web-UI fixes (Liam, voice 2026-06-14).

### 1. Smaller TTYD terminal font
The 4 elder terminals in the cockpit rendered at the xterm.js default of **15px**, so only a handful of lines fit per terminal. ttyd was launched with no font option, so it fell back to that default.

- **Fix:** `agents/entrypoint.sh` now launches ttyd with `-t fontSize=11`. `-t` / `--client-option` values pass straight through to the in-browser xterm.js terminal. 11px meaningfully increases visible rows/cols while staying readable.
- **Font size chosen:** `11px` (down from xterm.js default 15px), set in `agents/entrypoint.sh`.

> ⚠️ **Rebuild caveat:** this is a ttyd **server launch flag**, baked into the elder container entrypoint. It does **not** take effect on the running ttyd processes — it requires an **elder container rebuild/restart** to apply. No containers were restarted by this PR (out of scope per the worktree constraints).

### 2. Clan tab: human-readable ETA + action name
The clan tab (`ClansmanTab`) showed `ETA <raw number>` for in-progress actions. Now it shows the **action name** + a **human-readable local clock ETA**, e.g. **"Chop Wood … ETA 3:42:10 AM"**.

- The **action name** was already mapped server-side (`ACTION_LABEL_BY_ID` in `clansmen.ts`: Chop Wood, Mine Iron, Fish Docks, Harvest Wheat, …) and rendered as the row's primary label — so the action is shown alongside the ETA.
- **ETA:** the Convex query now derives an absolute settle wall-clock time (`etaTs`, unix seconds) by projecting the mission's `settlesAtTick` onto the tick epoch (`tickEpochStartedAt + settlesAtTick * tickDurationSeconds`). The client formats it as a local clock time with `toLocaleTimeString` (viewer's timezone). When `etaTs` can't be derived (tick epoch not yet surfaced), the tab falls back to the relative tick label (`ETA 2t`).
- Per Liam's note, this is a **stable absolute timestamp** (re-derives on each Convex query push) — **no client-side countdown/interval**, so there are no React timer/cleanup hazards.
- ETA grid column widened `60px → 88px` + `white-space: nowrap` so the clock time doesn't clip. Stub demo rows updated to seed realistic upcoming settle times.

## Files changed
- `agents/entrypoint.sh` — ttyd `-t fontSize=11`
- `apps/server/convex/clansmen.ts` — return `etaTs` (absolute settle unix seconds)
- `apps/web/src/components/cockpit/tabs/ClansmanTab.tsx` — render formatted ETA clock + widened column + stub seed

## Verification
- `pnpm --filter @clan-world/server typecheck` — **clean**
- `pnpm --filter @clan-world/web typecheck` — **clean**
- `bash -n agents/entrypoint.sh` — **clean**
- Server test suite: 126/127 pass. The 1 failure (`indexer.test.ts > preserves tick epoch start across same-tick snapshot refreshes`) is **pre-existing on `origin/dev`** (verified via stash) and untouched by this change.
- Note: `vite build` of `apps/web` fails with a pre-existing `[vite:build-html] URI malformed` error in `index.html` (also fails on clean `origin/dev`); the `tsc -b` typecheck stage passes.

## Screenshots needed
- Cockpit terminal tab after elder rebuild (to confirm the 11px font visually).
- Clan tab with an active mission, confirming "Chop Wood … ETA 3:42:10 AM" rendering + no column clipping.

Closes #675

### DIFF
--- a/agents/entrypoint.sh
+++ b/agents/entrypoint.sh
@@ -72,7 +72,14 @@ fi
 if [[ "${ADMIN_INJECT_ENABLED:-0}" != "1" ]]; then
   echo "[entrypoint] WARNING: ttyd is read-only and ADMIN_INJECT_ENABLED!=1. Operator input has NO channel. Set ADMIN_INJECT_ENABLED=1 once /api/admin/inject-message is deployed to prod." >&2
 fi
-ttyd --port "${TTYD_PORT}" tmux attach-session -t "${SESSION_NAME}" &
+# `-t fontSize=11` is an xterm.js client option (ttyd passes -t/--client-option
+# values straight through to the in-browser terminal). ttyd's default is 15px,
+# which only fits a handful of lines in each of the 4 cockpit elder terminals.
+# Dropping to 11px roughly doubles the visible rows/cols without becoming
+# unreadable. NOTE: this is a ttyd *server* launch flag — it only takes effect
+# after the elder containers are rebuilt/restarted (the running ttyd process
+# does not pick up the new flag).
+ttyd --port "${TTYD_PORT}" -t fontSize=11 tmux attach-session -t "${SESSION_NAME}" &
 TTYD_PID=$!
 echo "[entrypoint] ttyd started on port ${TTYD_PORT} (PID ${TTYD_PID})"
 
--- a/apps/server/convex/clansmen.test.ts
+++ b/apps/server/convex/clansmen.test.ts
@@ -1,0 +1,47 @@
+import { describe, expect, it } from "vitest";
+import { deriveEtaTs } from "./clansmen";
+
+// Mutation date: 2026-06-14. These tests pin the etaTs projection added in
+// PR #676 (super-swarm M2). Each case fails if the derivation is reverted:
+//  - drop the `tickEpochStartedAt <= 0` guard → the "missing epoch" case
+//    returns a relative-to-zero timestamp instead of null (FAILS).
+//  - drop the `eta === null` guard → the idle/dead case projects a bogus time
+//    instead of null (FAILS).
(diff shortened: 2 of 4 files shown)


claude-do:
**Orchestrator review synthesis — 1 MUST-FIX (full-context reviewer caught what the super-swarm missed):**

- **[MUST] `etaTs` formula is wrong** (`apps/server/convex/clansmen.ts`). `tickEpochStartedAt` is the wall-time of the **current** tick, not tick 0. Current `tickEpochStartedAt + settlesAtTick * tickDurationSeconds` makes every ETA hours off (e.g. settlesAtTick 502 @ 30s → ~4h future instead of ~60s). **Fix:** `tickEpochStartedAt + (settlesAtTick - currentTick) * tickDurationSeconds`. The relative-tick fallback (`ETA 2t`) is actually more correct than the current absolute time. (Note: the pre-existing `nowSeconds` cooldown calc has the same structural bug — out of scope but worth a follow-up.)
- **[SHOULD] `formatEtaClock` invalid-Date guard** (`ClansmanTab.tsx`): `new Date(NaN*1000).toLocaleTimeString()` renders the literal "Invalid Date" in the column. Guard with `isNaN(d.getTime())` → fallback string.

Super-swarm (5 models, 0 HIGH) MEDIUMs to also fold in: NOW_SECONDS module-load freeze (compute at runtime), missing etaTs test coverage, eta=0/past-clock render guard.

#claude-subagent-synthesis

Issue #652, opened 2026-06-13
heartbeat chain-clock anchor drifts vs anvil clock — periodic re-anchor needed (boot-static offset insufficient)


claude-do:
Follow-up to #649 / PR #651. The env-gated chain-clock scheduling measures the chain-vs-wall offset ONCE at boot. Observed live 2026-06-12 ~21:45 ET (4.5h after boot): runner intermittently fires one interval early and gets `heartbeat rate-limited` on roughly alternating attempts — 5 rate-limited vs 2 confirmed in a 6-min window — flapping the container healthcheck (webhook success-file aged past the 180s threshold). The mode self-corrects per-cycle (rate-limit handler re-reads next-allowed) so the game keeps ticking, but each cycle wastes reverted txs and the codex round-1 review explicitly flagged boot-static offset as the limitation.

Stopgap deployed on do-box: hourly `docker restart clan-world-heartbeat-1` via systemd user timer (`clanworld-heartbeat-restart.timer`) — each boot re-anchors. Bounds drift to ≤1h accumulation (measured drift ~0.16s/h, so well within jitter), at the cost of one delayed tick per hour.

Proper fix sketch: maintain a chain-clock estimator in heartbeatLoopMain instead of a frozen offset — poll `getBlock latest` every ~5-15s; when the block NUMBER changes, the new block is at most one poll-interval old, so snap the anchor to `block.timestamp` (slightly late-biased = safe, late never reverts); between snaps, project at wall pace. Every successful heartbeat mines a block, so the estimator re-anchors with ~seconds accuracy each cycle. Alternative: have the scheduler re-measure right after each confirmed tx (block age ≈ 0 at that instant) via an optional async dep. Either way, remove the restart timer once it lands.

Diagnostic note for the implementer: the alternating succeed/rate-limit log shape from the incident window is in do-box `docker logs clan-world-heartbeat-1` around 2026-06-13T02:00-02:06Z; fresh-boot offset measurements were 1487982ms (17:43 ET) vs 1488647ms (21:52 ET).

claude-do:
Update 22:00 ET: drift theory REFUTED as the primary cause. After a fresh container restart (clean re-anchor, offset 1488647ms), the alternating pattern persists: 4 confirmed / 3 rate-limited in the first 6 minutes. Measured anchor drift is only ~0.16s/h — far too small to explain firing a full interval early.

[...]

PR #677, opened 2026-06-14
fix(heartbeat): #652 — clean 30s cadence on anvil fork (r1–r5 + runner funding)

claude-do:
## Summary
Resolves the long-deferred #652 heartbeat flapping/freeze. Verified LIVE on a fresh anvil fork: **clean 30s cadence, container healthy, zero rate-limited reverts.**

## The real root cause (after 3 wrong theories)
The deepest cause of the frozen game was **the heartbeat runner wallet being out of ETH on the fork** (drained to 0.000234 ETH by an earlier revert flood — every reverted tx still burns gas). Diagnostic signature: `cast call heartbeat()` from the runner SUCCEEDS (eth_call needs no gas) while the runner's real tx reverts with **empty data** (`0x`) = insufficient funds. Fixed live via `anvil_setBalance` (1000 ETH) — and this must become a permanent fork-bootstrap step.

## The fix stack (commits)
- **r1** `c98602b` — safety margin (no early fire) + treat rate-limited revert as wait-in-window, not a back-off-and-miss.
- **r2** `0c7a88f` — fork-time advance (`evm_setNextBlockTimestamp`/`evm_increaseTime` + `evm_mine`), **double-gated to dev-fork only** (`CHAIN_NETWORK=dev` AND local-fork RPC, or explicit `HEARTBEAT_ADVANCE_FORK_TIME=1`) — never runs against Base Sepolia. Retry floor (no 0ms hot-loop). Telegram-token-absent → no-op.
- **r3** `fba8a85` — send-path diagnostics (chainBlockTs/nextHeartbeatAtTs/wallTs/gasLimit) so empty reverts are legible; dev gas default.
- **r4** `e423d97` — wall-pace cadence in fork-advance mode (chain-clock scheduling was making every cycle "due").
- **r5** `d6a984d` — pace by **wall interval since last beat** (not absolute chain-time target), fixing a 63-min sleep when the fork clock had diverged ahead of wall.

## Tests + live verification
- 76/76 heartbeat package tests pass; typecheck clean. New tests model the fork case (chain clock diverged from wall) — the gap that let earlier green tests ship a broken live game.
- Live: fresh fork → 3 confirms in 95s, `computedDelayMs ≈ 28990` (waiting ~29s between beats), healthy.

## Follow-ups (tracked separately)
- Make runner-funding a permanent fork-bootstrap step (fund `0xBC34…6cc7` on every fresh fork).
- Consider lowering the r3 25M dev-gas default now that the wallet is funded.

Closes #652

### DIFF
--- a/packages/heartbeat/.env.example
+++ b/packages/heartbeat/.env.example
@@ -32,6 +32,11 @@ RUNNER_ID=clanworld-runner
 # Optional heartbeat success marker path override; defaults to /tmp/last-heartbeat-success.
 # HEARTBEAT_SUCCESS_FILE_OVERRIDE=
 
+# DEV/FORK ONLY — never set these against a real chain.
+# CHAIN_NETWORK=dev
+# HEARTBEAT_ADVANCE_FORK_TIME=1
+# HEARTBEAT_SCHEDULE_FROM_CHAIN=1
+
 # Optional Telegram alerting for heartbeat retry exhaustion. If the bot token is
 # missing or Telegram fails, the runner logs locally and keeps running.
 TELEGRAM_BOT_TOKEN=
--- a/packages/heartbeat/README.md
+++ b/packages/heartbeat/README.md
@@ -37,7 +37,7 @@ Heartbeat cadence comes from the diamond's owner-configured
 2. Restart the runner or `scripts/start-heartbeat-loop.sh`.
 
 The scheduler reads `heartbeatIntervalSeconds()` once at boot, then schedules
-each fire from `getWorldState().nextHeartbeatAtTs` with a 500 ms jitter buffer.
+each fire from `getWorldState().nextHeartbeatAtTs` with a 1,500 ms safety margin.
 
 ## Alerts and status
 
--- a/packages/heartbeat/src/heartbeatScheduler.ts
+++ b/packages/heartbeat/src/heartbeatScheduler.ts
@@ -69,6 +69,14 @@ export function computeHeartbeatDelayMs(
   return Math.max(0, nextHeartbeatAtTs * 1000 + safetyMarginMs - nowMs);
 }
 
+export function computeWallIntervalDelayMs(
+  intervalMs: number,
+  lastBeatWallMs: number,
+  wallNowMs: number,
+): number {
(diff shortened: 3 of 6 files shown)


claude-do:
**Gemini flash review** · round 1 · head `d6a984d2`

Verdict: CLEAN

No high or medium findings. Two low-severity informational notes:

- [LOW] heartbeatScheduler.ts:32 — `HEARTBEAT_JITTER_MS` re-exported as alias for `HEARTBEAT_SAFETY_MARGIN_MS`; name no longer reflects deterministic vs. random behavior (non-breaking, informational only)
- [LOW] runnerCastHeartbeat.ts:157 — `Date.now()` fallback in `isNextHeartbeatStillFuture` if `getBlock` fails is safe but untested path

Production safety gate (`CHAIN_NETWORK === 'dev'` + local RPC URL check) verified present. Rate-limit detection uses block timestamp not wall clock — correct. Retry budget separation confirmed.

#gemini-flash

claude-do:
**Codex review** · round 1 · head `d6a984d`

Verdict: NEEDS_FIXES (1 HIGH, 2 MEDIUM)

- [HIGH]  packages/heartbeat/src/runnerCastHeartbeat.ts:424 — `HEARTBEAT_ADVANCE_FORK_TIME=1` bypasses the local-fork RPC gate. If set alongside a Sepolia RPC, `advanceForkTimeTo()` fires `evm_setNextBlockTimestamp`/`evm_mine` against the real chain. The override must still enforce a local-fork RPC check.
- [MED]   packages/heartbeat/src/heartbeatScheduler.ts:263 — rate-limit retry branch never increments `attempt` and has no max retry count; can trap scheduler in `attemptHeartbeatWithBackoff()` indefinitely (no budget consumed, outer-cycle alerts suppressed).
- [MED]   packages/heartbeat/src/heartbeatScheduler.ts:273 — rate-limit wait delay uses `computeHeartbeatDelayMs()` (chain-time based), not wall cadence. In fork mode post-advance, collapses to safety-margin floor only, hammering fork every ~2s on repeated rate-limit reverts — conflicts with wall-interval pacing.

#codex

